PR #56, opened 2018-09-24
relayer-identityService: remove pending authorisation 

JustynaBroniszewska:
Remove pending authorisation from authorisation service if execute add key

### DIFF
--- a/universal-login-relayer/lib/services/IdentityService.js
+++ b/universal-login-relayer/lib/services/IdentityService.js
@@ -1,14 +1,16 @@
 import Identity from '../../build/Identity';
 import {addressToBytes32} from '../utils/utils';
-import ethers, {utils} from 'ethers';
+import ethers, {utils, Interface} from 'ethers';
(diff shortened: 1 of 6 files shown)


universal-login-relayer/lib/services/IdentityService.js
```diff
@@ -27,6 +29,12 @@ class IdentityService {
 
   async executeSigned(contractAddress, message) {
     const contract = new ethers.Contract(contractAddress, this.abi, this.wallet);
+    const addKeySighash = new Interface(Identity.interface).functions.addKey.sighash;
+    if (message.to === contractAddress && message.data.slice(0, addKeySighash.length) === addKeySighash) {
+      const [address] = (this.codec.decode(['bytes32', 'uint256', 'uint256'], message.data.replace(addKeySighash.slice(2), '')));
+      const key = utils.hexlify(utils.stripZeros(address));
```
marekkirejczyk: Two lines above could be easily extracted to function, eg `const [key, address] = decodeData(addKeySighash);`

PR #65, opened 2018-09-25
Remove DApp specific contract from contracts dir

Kyrrui:
The clicker.sol contract should not exist in the universal login contracts dir -- it should be in the example dir since they are conceptually disparate projects.


marekkirejczyk:
LGTM Thank you :) ❤️ 👍

PR #68, opened 2018-09-27
Standardized blockies

alexvandesande:
fixed size and setting of blocks into ethereum standard, also added on history view


alexvandesande:
screen shot 2018-09-28 at 9 25 43 am


There's one issue I was not able to fix on this branch: on the history view, the blockies are not consistent with the main one. This is because in that prop, event.address doesn't return the actual address but the name, so the blockie ends up using the ens name as a seed, which is wrong. The prop should have both `event.address` and `event.name` as separate items

screen shot 2018-09-28 at 9 25 56 am

PR #69, opened 2018-09-27
Decrement number of tokens on a click

Kyrrui:


marekkirejczyk:
Hi @Kyrrui. Thank you for the PR.
It is implementing a feature that we work on now but in a bit different way.
The way we would like it to work is the tokens are being used to pay for a transaction fee.
The relayer will pay the fee in Ether, while Smart Contract will pay tokens back to relayer.

I am closing for now. We are looking to your futher contributions in the future!
Feel free to get in touch if there is anything, in particular, you would like to work on.

Kyrrui:
@marekkirejczyk Could you please loop me into your task board so that I can work in parallel :)

PR #75, opened 2018-09-28
Simplified Login

alexvandesande:
This is an on going work branch that I wanted to share as it will end up changing a lot of things on the login side BUT I believe it will be for the better 

screen recording 2018-09-28 at 11 19 14 am

The app starts in a much cleaner state, which allows much greater flexibility for each app developer to give their own style and makes the SDK more universal. Then as the user types a name the app shows a hover wich will show the option to either login or create other names in other domains. If the name exists it will also suggest a recovery.

Not implemented: 

* key up, key down and enter support.
* all domains except the top ones are faked
* recovery option is visible but not working


alexvandesande:
@marekkirejczyk I pushed this branch as far as I could. I would like you and justyna to take a look and see if it can be merged. I think it only needs support for common keys (enter, up, down). Support for other domains or recovery is not supported either, but I don't believe it is anywhere so it doesn't matter.

alexvandesande:
screen recording 2018-10-01 at 1 09 28 pm

Made some more improvements and (maybe unnecessary) animations. 😃

PR #82, opened 2018-10-02
Generating signed message with payment options. Identity token refund.

JustynaBroniszewska:


universal-login-contracts/contracts/Clicker.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.4.24;
+
+
+/* solium-disable security/no-block-members*/
+contract Clicker {
```
Kyrrui: I don't think Clicker.sol should be a part of the universal-login-contracts directory. It is a contract related to the example application and should live there. :)

alexvandesande: We first want the app to be working, ideally all the bits: the example, the relayer code, the sdk, should live in different repositories altogether. For now, leave it be

marekkirejczyk: I think Justyna used in tests. I think we can move it easily back to example and use mock contract for test.

universal-login-contracts/contracts/Identity.sol
```diff
@@ -5,8 +5,10 @@ import "./ENSRegistered.sol";
 
 
 contract Identity is KeyHolder, ENSRegistered {
-    constructor(bytes32 _key, bytes32 _hashLabel, string _name, bytes32 _node, ENS ens, FIFSRegistrar registrar, PublicResolver resolver) payable public
+    constructor(
+        bytes32 _key, bytes32 _hashLabel, string _name, bytes32 _node, ENS ens, FIFSRegistrar registrar, PublicResolver resolver, address relayerAddress) 
+        payable public
         ENSRegistered(_hashLabel, _name, _node, ens, registrar, resolver)
-        KeyHolder(_key) {
+        KeyHolder(_key, relayerAddress) {
```
marekkirejczyk: I would propose to send tokens back to msg.sender when calling exectuteSigned instead of setting relayer in the constructor.

universal-login-contracts/contracts/KeyHolder.sol
```diff
@@ -201,13 +211,46 @@ contract KeyHolder is ERC725 {
...
             emit Executed(_id, executions[_id].to, executions[_id].value, executions[_id].data);
         } else {
             emit ExecutionFailed(_id, executions[_id].to, executions[_id].value, executions[_id].data);
         }
     }
+
+    function refund(uint256 _id, uint256 _gasUsed) private {
+        if (executions[_id].gasToken != address(0)) {
+            StandardToken token = StandardToken(executions[_id].gasToken);
+            token.transfer(relayerAddress, _gasUsed.mul(executions[_id].gasPrice));
```
marekkirejczyk: I would use here `msg.sender` and get rid of `relayerAddress` completely.

universal-login-contracts/test/contracts/KeyHolder/Executions.js
```diff
@@ -204,7 +207,21 @@ describe('Key holder: executions', async () => {
...
+    describe('Refund', async () => {
+      it('Should refund after execute', async () => {
+        const gasPrice = utils.parseEther('0.00011').toString();
+        const relayerTokenBalance = utils.formatEther(await mockToken.balanceOf(wallet.address));
+        signature = messageSignature(
+          wallet, targetAddress, identity.address, amount, data, 0, mockToken.address, gasPrice, 1);
+
+        await mockToken.transfer(identity.address, utils.parseEther('20'));
+        await identity.executeSigned(targetAddress, amount, data, 0, mockToken.address, gasPrice, 1, signature);
+        expect(utils.formatEther(await mockToken.balanceOf(wallet.address))).not.to.eq(relayerTokenBalance);
```
marekkirejczyk: extract variable: 
`expectedBalance = utils.formatEther(await mockToken.balanceOf(wallet.address))`

PR #83, opened 2018-10-02
Mnemonic alternative

Kyrrui:
This is an alternative to the Daefen implementation -- since Daefen is not yet a npm library and there is no standard for converting to/from private keys with Daefen I thought we might want to consider the more standard mnemonic route.
 (This is an alternative to https://github.com/EthWorks/UniversalLoginSDK/pull/81)


alexvandesande:
> no standard for converting to/from private keys

Yes there is and it's implemented in ether.js it's called Brainwallet

No, I am open to any other standard for generating high entropy short pseudo words that you propose, but I would NOT like to use a 12 word seed standard

Kyrrui:
Brainwallet was deprecated from ether.js unfortunately

alexvandesande:
Mmm I see 😞 

That's too bad, because the issue with Brain wallets is the question of picking a password with high enough entropy. It's a question of user experience: 12 word seeds are longer and harder to type, and therefore more suitable for HD seeds, not individual accounts. Ether.js wordlist has 2048 words, so 12 words is about 132 bits of entropy, which is the equivalent of 11 syllables in Daefen (""Liwkef Kutnud Sihaco Romjur Ellyt Nue").

screen shot 2018-10-03 at 9 28 11 am
(who would type that?!)

 I think that is still a lot for a user to type I would think a 8 syllable / 4 words phrase ("Sahbyg Feufoo Azsak Esfod"), that still has about 94 bits of entropy to be sufficient. The way that would be generated would be simply hash username+seed N times and then use the result as the key. Just to give an idea, all ethereum miners do, currently, 300 petahashes per second, so in order to break a single of these keys you'd have to use the whole network power for 2 thousand years, for every N we choose (which can be in thousands, since I suggest we take about 10 seconds to generate it on a consumer device). 

Considering that this is for one account, and that we want to give them limited powers by contract so that one lost account is not the end of the world, I would say that this is a better UX/security approach.

This is a hill I'm willing to die on, specially because at the moment Universal Logins is a *user experience demonstration*, and I want to show users we can do better in multiple fronts.

Issue #76, opened 2018-09-30
Generating backup keys


marekkirejczyk:
We would like to implement the Backup Keys functionality in the example application.

Functionality is mocked in the application already.
To get there one needs to login/create identity than click Account -> Backup code -> "GENERATE NEW CODE".

The feature is pretty big, and therefore I would propose to split it into multiple PR:

### 1.  UI and backup phrase generation
This will require changes in React `Backup` and ``BackupView components in `universal-login-example`.

- [ ] Generate and display three backup codes when screen shows-up
- [ ] Display actually Identity info (should be taken from `identityService`, which should be passed as props to `Backup` from `ContentContainer`.
- [ ] Implements 'generate 3 more codes' button
- [ ] Use [Daefen](https://github.com/alexvandesande/Daefen)

Even this feature is quite big, so could split into 2-4 PRs.

### 2. Implement adding multiple keys in smart contracts
That will require adding a similar method to 
```solidity
function addKey(bytes32 _key, uint256 _purpose, uint256 _type)
```
I propose the following format: 
```solidity
function addKeys(bytes32 [] _key, uint256 [] _purpose, uint256 [] _type)
```
Function would simply iterate ove arguments and call addKey.

Kyrrui:
I will take on this task :)!

Kyrrui:
Today I am working on step 1 of this issue "1. UI and backup phrase generation"

Kyrrui:
@marekkirejczyk Perhaps a step 4 should be added to this ticket once adding multiple keys has been completed - 
Step 4. Create a alternative login where user uses backup code

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 350.0 DAI (350.0 USD @ $1.0/DAI)  attached to it__ as part of the Ethereum Community Fund via ECF Web 3.0 Infrastructure Fund  fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/EthWorks/UniversalLoginSDK/76/1348).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/EthWorks/UniversalLoginSDK/76/1348).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
[...]

PR #85, opened 2018-10-03
[#76] Full account recovery with backup codes.

Kyrrui:
This is the full account recovery flow hooked together. 

To test:
1. Sign into an account
2. Generate backup codes
3. Copy one of the backup codes
4. Attempt log into alternative device
5. Enter one-time code you copied.

The UI should be changed for this (currently the button labels change and nothing else @alexvandesande  perhaps you can help with this.)

I'm going to leave the other pull request up so that if we decide on a different flow for hooking everything together we can work with that branch.

**Question**
Are backup codes one time use? If yes I suggest:
1. Adding a keySwap method to the keyHolder.sol which adds and removes a key from the contract in one transaction (Adds device key, removes recovery key)
2. Making it clearer to the user that these are one time codes.


alexvandesande:
It's very good and seems to be working. I would wait until the simple-login and token-buyback is merged before we can move forward with most of the UI changes. I will push marek on this. Meanwhile, here's my feedback:

1. When you go to the backup option, "GENERATE 3 MORE CODES" button should **append**, **not replace** the previous code. It's for someone who wants to generate 6, 9 codes, etc. Having such important codes just disappear with a button feels flimsy.

screen shot 2018-10-03 at 5 12 55 pm

2. Once you generate it, the "successful" screen is confusing: what should you do? clicking "success" generates new codes. What happens to the previous ones? You need to click "cancel" there. **my suggestion: when successful, move back to the account settings page**

3. The settings page does not reflect anything, so it might be unclear what happened to the recovery accounts. This can be solved by implementing the "manage devices" accordion. What would happen is that clicking them it would list all the current devices and recovery accounts, and on the backup option it should be clearly stated "you've done this already but you can do it again":

screen shot 2018-10-03 at 5 25 04 pm

Now, this creates an issue, which is that we don't have any meta information about these accounts, so a new device won't be able to load the "names" from anywhere. We could add that information on the contract but that would be not only expensive but also leak private data (like which devices and OS you use). We can have more advanced ways in which we sign data and send it to whisper, but I'd leave that for later and for now I would simply have locally saved names (maybe with inline editing?) so that a new device would simply see all accounts as 1, 2, 3, etc.

That's my feedback for now. I have more on the right user interface, but that will need to wait for all the merges with simple login stuff..

Kyrrui:
@alexvandesande 
1. Oh I see, I was unaware that we wanted to append to the list of backup codes -- I can make this change quickly and add to this PR.
2. Easy change, I will have it re-direct to the account settings page.
3. We will continue this discussion in https://github.com/EthWorks/UniversalLoginSDK/issues/86

I'm wondering -- these are one time codes correct? Currently I implement them as such.

alexvandesande:
> I'm wondering -- these are one time codes correct? Currently I implement them as such.

What do you mean by that? On the backend they are just indistinguishable from other accounts and could be reused multiple times even as a standard login system if the user so wanted. But we don't want to encourage that behavior, instead we want to encourage that people (1) generate them, (2) print them, (3) store them safely, (4) use them in case they need one.

One thing we could do then, is after successfully recovering we could show a message encouraging the person to throw away that paper and print more recovery codes. What do you think?

alexvandesande:
My backup keys don't take any time at all to generate, I never even see the "loading screen". Is it doing it correctly? Is my machine simply faster?

universal-login-example/src/components/ApproveConnection.js
```diff
@@ -22,10 +27,26 @@ class ApproveConnection extends Component {
     await sdk.denyRequest(identityAddress, address);
   }
 
+  async onRecoverClick() {
+    this.setState({btnLabel: 'Recovering...'});
+    const {identityService, sdk} = this.props.services;
+    let wallet = await Wallet.fromBrainWallet(this.identityService.identity.name, this.state.backupCode);
+    // TODO: We should probably write into the contract a 'Swap' function which adds a key and removes another key in one transaction.
+    await sdk.addKey(identityService.identity.address, identityService.deviceAddress, wallet.privateKey, DEFAULT_PAYMENT_OPTIONS);
+    await sdk.removeKey(identityService.identity.address, wallet.address, wallet.privateKey, DEFAULT_PAYMENT_OPTIONS);
```
alexvandesande: Oh I see what you meant by "one time use only". This is actually an interesting idea, didn't consider using the key exclusively to login.

There's an issue however: in some cases, adding and removing a key, might require more than one confirmation for security purposes, meaning that you might be in a situation where you're using a key to recover but might need the lost keys to approve it.

In order to simplify it, I would simply encourage users to throw away the recovery paper and create more recovery keys in the future as an optional step.

universal-login-example/src/components/Backup.js
```diff
@@ -1,14 +1,74 @@
...
+import ethers from 'ethers';
+import DEFAULT_PAYMENT_OPTIONS from '../../config/defaultPaymentOptions';
 
 class Backup extends Component {
+  constructor(props) {
+    super(props);
+    this.identityService = this.props.services.identityService;
+    this.sdk = this.props.services.sdk;
+    this.state = {
+      backupCodes: ['loading...', 'loading...', 'loading...'],
```
alexvandesande: I never see those. It feels "too fast" for me. Is it just a question of my machine?

universal-login-example/src/components/Backup.js
```diff
@@ -1,14 +1,74 @@
...
+    this.setState({ btnLabel: 'Setting...' });
+    const {identityService} = this.props.services;
+    const to = identityService.identity.address;
+    const {privateKey} = identityService.identity;
+    const {sdk} = identityService;
+    var wallets = await Promise.all([
+      ethers.Wallet.fromBrainWallet(identityService.identity.name, this.state.backupCodes[0]), 
+      ethers.Wallet.fromBrainWallet(identityService.identity.name, this.state.backupCodes[1]), 
+      ethers.Wallet.fromBrainWallet(identityService.identity.name, this.state.backupCodes[2])
+    ]);
```
alexvandesande: I would generate these from the backup code before hand and only show each backup once the key itself was generated. This gives the user a more permanent feeling of "working" to generate these and makes so that the set up process itself is faster. Because it's slower, it gives the user more information on the drawbacks if they want to generate 6 or more.

universal-login-example/src/components/Backup.js
```diff
@@ -1,14 +1,74 @@
...
   render() {
-    return <BackupView setView={this.props.setView} />;
+    const { identity } = this.props.services.identityService;
+    return (
+      <BackupView
+        btnLabel={this.state.btnLabel}
+        identity={identity}
+        setView={this.props.setView}
+        backupCodes={this.state.backupCodes}
+        onGenerateClick={this.generateBackupCodes.bind(this)}
```
alexvandesande: Clicking on generating the button should **append** new code on the end of the queue, not replace them

universal-login-example/src/views/BackupView.js
```diff
@@ -8,30 +8,60 @@ class BackupView extends Component {
...
-            your username as it&apos;s required to recover access. These are
-            independent codes, use each one once. Keep offline and away from
-            computers.
+            Print these, cut them apart and keep them safe locations apart from
+            each other. Keep them away from computers until you want to use
+            them.
           </p>
           <hr className="separator-s" />
-          <p className="backup-code bold">bamdaa-ewar-izoisi</p>
+          <div className="row align-items-center">
```
alexvandesande: we should not hard code 3 items, instead we should have an unbounded array and add 3 more at each time the user clicks to generate more codes

universal-login-example/src/components/ApproveConnection.js
alexvandesande: This would probably require a new screen, similar to the greetings screen, or something else, telling the user that. Something similar in layout to other screens we already have:

screen shot 2018-10-04 at 10 40 26 am

universal-login-example/src/components/Backup.js
alexvandesande: Oh I see what is happening! More below

universal-login-example/src/components/ApproveConnection.js
```diff
@@ -2,12 +2,17 @@ import ApproveConnectionView from '../views/ApproveConnectionView';
...
+import DEFAULT_PAYMENT_OPTIONS from '../../config/defaultPaymentOptions';
 
 class ApproveConnection extends Component {
   constructor(props) {
     super(props);
     this.sdk = this.props.services.sdk;
     this.identityService = this.props.services.identityService;
+    this.state = {
+      backupCode: '',
+      btnLabel: 'RECOVER ACCOUNT'
```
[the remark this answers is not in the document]
Kyrrui: Where should it live? Could it be a link off of the requesting connection screen?

universal-login-example/src/components/ApproveConnection.js
Kyrrui: I will make these changes, but I will say I think having a bunch of keys accumulate on an identity is dangerous for security. That means these sheets of paper are free rides into an account permanently unless the user deletes them.

universal-login-example/src/components/Backup.js
Kyrrui: Will make this change

universal-login-example/src/components/Backup.js
Kyrrui: Will make this change

universal-login-example/src/views/BackupView.js
Kyrrui: Will make this change

universal-login-example/src/components/ApproveConnection.js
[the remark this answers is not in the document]
alexvandesande: It already exists on that screen. When you type an existing name it gives you two options: connect or recover

screen shot 2018-10-04 at 12 05 44 pm

But since some users may not find it, we should probably also add a link from the connection requested screen

universal-login-example/src/components/ApproveConnection.js
alexvandesande: Which is why the user should probably burn the little pieces of paper 😉 
But then, they might be useful as recoveries. I think that you'll be correct long term and that the safer way will be to replace the keys when used the first time, but I'd wait until we have the whole multi-approve process smoothed out

universal-login-example/src/components/ApproveConnection.js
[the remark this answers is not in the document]
Kyrrui: Okay, I'll hook up a link from the connection requested screen for now and then will try my hand at implementing this.

Kyrrui:
@alexvandesande @marekkirejczyk  I implemented all the requested changes -- but there is one minor concern. If you generate 9+ codes and hit backup, an error is thrown - I believe this to be an issue with the default gas settings but I'm not sure.

alexvandesande:
The only small thing that bothers me is that when you click "set as backup code" there's really no feedback that it was saved. My suggestion would be:

1. More clearly communicate the status of the backup codes. The account accordion always says "you have saved this" even when you haven't. We definitely need two layouts for that accordion and its internal buttons to show the current state of the recovery code

2. Implement the "manage devices" accordion, where it will show you a list of devices and accounts with their names. In this case it might be worth it when you come back from the backup screen, to even open the accordion to show the new listed accounts.

universal-login-example/src/Daefen.js
```diff
@@ -0,0 +1,123 @@
+/*
```
marekkirejczyk: Why not add as npm?

Kyrrui: wasn’t sure of the proper way of adding this... It’s not an npm library yet

Kyrrui: wasn’t sure of the proper way of adding this... It’s not an npm library yet

Kyrrui:
> The only small thing that bothers me is that when you click "set as backup code" there's really no feedback that it was saved. My suggestion would be:
> 
> 1. More clearly communicate the status of the backup codes. The account accordion always says "you have saved this" even when you haven't. We definitely need two layouts for that accordion and its internal buttons to show the current state of the recovery code
> 2. Implement the "manage devices" accordion, where it will show you a list of devices and accounts with their names. In this case it might be worth it when you come back from the backup screen, to even open the accordion to show the new listed accounts.

@alexvandesande See if you like what I did with #1, as for #2 I think we should add that in a different pull request (this one is already quite large XD). Cheers!

universal-login-example/src/Daefen.js
marekkirejczyk: I that case let's move it to `src/vendors` and add that dire to linter ignore

universal-login-example/src/components/ApproveConnection.js
```diff
@@ -1,32 +1,51 @@
 import ApproveConnectionView from '../views/ApproveConnectionView';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import {Wallet} from 'ethers';
+import { Wallet } from 'ethers';
```
marekkirejczyk: We should probably agree on guidelines regarding spaces in `{}` and put it to linter, otherwise, it generates noise in the repository.

universal-login-example/src/components/Account.js
```diff
@@ -28,14 +27,15 @@ class Account extends Component {
...
           <hr className="separator" />
-          <ManageDevicesAccordion emitter={this.props.identityService.emitter} />
+          <ManageDevicesAccordion
+            emitter={this.props.identityService.emitter}
+          />
           <hr className="separator" />
           <BackupCodeAccordionView setView={this.setView.bind(this)} />
           <hr className="separator" />
-          <TrustedFriendsAccordionView setView={this.setView.bind(this)} />
-          <hr className="separator" />
```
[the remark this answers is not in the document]
alexvandesande: Because it's not in scope, no one is working to implement it and it adds complexity to it. If someone has actual plans to do it, then we could put it back

universal-login-example/src/components/ApproveConnection.js
alexvandesande: Sorry, it's my auto formatter

alexvandesande:
> If you generate 9+ codes and hit backup, an error is thrown - I believe this to be an issue with the default gas settings but I'm not sure.

In this case, maybe we could limit to 9 codes max?

> See if you like what I did with 1, as for 2 I think we should add that in a different pull request (this one is already quite large XD). Cheers!

I like the modal. We'll need for other stuff too, so it's useful. Also happy that it allows a print option as I was missing that. I would change a bit the layout here and there, but I can do it in a different PR in the future. 

LGTM!

universal-login-example/src/components/Backup.js
```diff
@@ -1,14 +1,80 @@
...
+      publicKeys: [],
+      isLoading: false
+    };
+  }
+
+  componentDidMount() {
+    this.generateBackupCodes();
+  }
+
+  async generateBackupCodes() {
```
marekkirejczyk: I would extract this one to a separate class, let's say `BackupService`, that would have one key method `generateBackupCodes()` that would return a pair of arrays (publicKeys, code). So that component will mount would look like following:
```js
componentDidMount() {
  const [backupCodes, publicKeys] = this.props.backupService.generateBackupCodes();
  this.setState({ backupCodes, publicKeys});
}

File: universal-login-example/src/components/RecoverAccount.js [hunk 11]
```diff
@@ -0,0 +1,61 @@
...
+    const {address} = new Wallet(identityService.privateKey);
+    const {sdk} = identityService;
+    await sdk.denyRequest(identityAddress, address);
+  }
+
+  async onRecoverClick() {
+    this.setState({isLoading: true});
+    const {identityService, sdk} = this.props.services;
+    let wallet = await Wallet.fromBrainWallet(this.identityService.identity.name, this.state.backupCode);
+    await sdk.addKey(identityService.identity.address, identityService.deviceAddress, wallet.privateKey, DEFAULT_PAYMENT_OPTIONS);
```
alexvandesande: There's no error handling here. When you input an incorrect recovery code, the app just keeps spinning forever

universal-login-example/src/views/Modals/BackupModal.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+class BackupModal extends Component {
```
marekkirejczyk: I think it should be called `BackupModalView`. While there should be perhaps other component `BackupModal` that uses `BackupModalView`.

universal-login-sdk/lib/sdk.js
```diff
@@ -48,6 +48,29 @@ class EthereumIdentitySDK {
...
+    const message = {
+      to,
+      from: to,
+      value: 0,
+      data,
+      nonce: transactionDetails.nonce,
+      gasToken: transactionDetails.gasToken,
+      gasPrice: transactionDetails.gasPrice,
+      gasLimit: transactionDetails.gasLimit
+    };
```
marekkirejczyk: Shorter version:
```js
    const message = {
      ...transactionDetails
      to,
      from: to,
      value: 0,
      data
    };
```

universal-login-sdk/lib/sdk.js
```diff
@@ -48,6 +48,29 @@ class EthereumIdentitySDK {
...
 
+  async addKeys(to, publicKeys, privateKey, transactionDetails) {
+    const keys = [];
+    const keyRoles = [];
+    const keyTypes = [];
+    for (const publicKey of publicKeys) {
+      keys.push(addressToBytes32(publicKey));
+      keyRoles.push(MANAGEMENT_KEY);
+      keyTypes.push(ECDSA_TYPE);
+    }
```
marekkirejczyk: Shorter version:
```js
    const keys = publicKeys.map((publicKey) => addressToBytes32(publicKey));
    const keyRoles = new Array(publicKeys.length).fill(MANAGEMENT_KEY);
    const keyTypes = new Array(publicKeys.length).fill(ECDSA_TYPE);
```

universal-login-sdk/test/sdk.js
```diff
@@ -103,6 +104,18 @@ describe('SDK - integration', async () => {
...
 
+    describe('Add keys', async () => {
+      const transactionDetalis = {
+        nonce: 0,
+        gasToken,
+        gasPrice,
+        gasLimit
+      };
+      it('should return execution nonce', async () => {
+        expect(await sdk.addKeys(identityAddress, [otherWallet.address, otherWallet2.address], privateKey, transactionDetalis)).to.eq(4);
```
marekkirejczyk: Perhaps a check if keys are added could be useful.

marekkirejczyk:
Looks good! Added some proposal for minor improvements.  👍 ❤️ ❤️ ❤️ 
Can't wait to merge it!

Kyrrui:
> Looks good! Added some proposal for minor improvements. 👍 ❤️ ❤️ ❤️
> Can't wait to merge it!

Had to go to some meetings this morning, but am working on these improvements now -- should be ready to merge by EoD

Kyrrui:
@marekkirejczyk @alexvandesande 
I think it's all set to go, the only two things that may want to be added/changed that I couldn't figure out:
1. When you generate 5 codes, I hide the generate button, but for some reason this messes up the layout of the `set backup codes` button on the page.
2. I was unable to figure out how to use the BlockchainRelayer in `test/sdk.js` to test adding multiple keys.

universal-login-example/src/services/BackupService.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  clearBackupCodes() {
+    this.backupCodes = [];
+    this.publicKeys = [];
+  }
+
+  async generateBackupCodes(numCodes) {
+    for (var i = 0; i < numCodes; i++) {
+      var backupCode =
```
marekkirejczyk: const

universal-login-example/src/services/BackupService.js
```diff
@@ -0,0 +1,37 @@
...
+        toWords(Math.floor(Math.random() * Math.pow(3456, 4)))
+          .replace(/\s/g, '-')
+          .toLowerCase() +
+        '-' +
+        toWords(Math.floor(Math.random() * Math.pow(3456, 4)))
+          .replace(/\s/g, '-')
+          .toLowerCase();
+      var wallet = await ethers.Wallet.fromBrainWallet(
```
marekkirejczyk: const

universal-login-example/src/services/BackupService.js
```diff
@@ -0,0 +1,37 @@
...
+        this.identityService.identity.name,
+        backupCode
+      );
+      this.backupCodes.push(backupCode);
+      this.publicKeys.push(wallet.address);
+    }
+    return [this.backupCodes, this.publicKeys];
+  }
+}
```
marekkirejczyk: Would be good to add a test.

universal-login-example/src/components/Modals.js
```diff
@@ -23,12 +24,25 @@ class Modals extends Component {
     this.props.emitter.emit('showModal', null);
   }
 
+  showAccount() {
+    this.hideModal();
+    this.props.emitter.emit('setView', 'Account');
+  }
+
+  async printScreen() {
```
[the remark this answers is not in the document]
marekkirejczyk: Reviewed ✅

universal-login-example/src/components/RecoverAccount.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  async onRecoverClick() {
+    this.setState({isLoading: true, msg: ''});
+    const {identityService, sdk} = this.props.services;
+    let wallet = await Wallet.fromBrainWallet(this.identityService.identity.name, this.state.backupCode);
+    try {
+      await sdk.addKey(identityService.identity.address, identityService.deviceAddress, wallet.privateKey, DEFAULT_PAYMENT_OPTIONS);
+    } catch (error) {
+      this.setState({isLoading: false, msg: 'Incorrect backup code, please retry'});
```
marekkirejczyk: `msg` -> `message`

universal-login-example/src/components/RecoverAccount.js
marekkirejczyk: Reviewed ✅

universal-login-example/src/views/Modals/BackupModal.js
marekkirejczyk: Reviewed ✅

universal-login-sdk/lib/sdk.js
marekkirejczyk: Reviewed ✅

universal-login-sdk/test/sdk.js
marekkirejczyk: I see there is no check in `Add key` as well. This SDK test, it is covered in contracts test, so I think it fine. 
Reviewed ✅

marekkirejczyk:
Looks good. There are four more minor improvements, but I think it is fine to do them in a separate PR.
Merging. 

Good job! 👍 👍 👍

universal-login-example/src/services/BackupService.js
Kyrrui: ✔️

universal-login-example/src/services/BackupService.js
Kyrrui: ✔️

universal-login-example/src/components/RecoverAccount.js
Kyrrui: ✔️

PR #101, opened 2018-10-10
README splitted into 3 separate README`s in 3 catalogs

JustynaBroniszewska:


README.md
```diff
@@ -12,208 +12,14 @@ Planned functionality for first release include:
...
-```js
-{
-  privateKey: '0x123...123',
-  port: 3311
-}
-```
-
-Default port is `3311`, `privateKey` is the only required field.
-
-## Running example on dev environment (quick option)
```
alexvandesande: I would keep this section on the main readme, as it tells a new user in 3 command lines how to run the demo

Issue #93, opened 2018-10-05
Save private keys in non-volatile areas


alexvandesande:
Currently, the private keys are stored in session storage and are lost when the app refreshes. While that's actually useful for testing purposes for fast login, we need to move all the app information to a more secure place: localstorage, localdb, cookies, maybe even in more than one place, so that the keys are hard to lose and sessions persists.

Kyrrui:
I will take on this task.

I have implemented this in the boomerang fork of this project for just local storage, but I would like to research this topic more and develop a more secure solution

Kyrrui:
After doing some research, [this](https://developer.chrome.com/apps/offline_storage#persistent) seems like the best option. I will look for a wrapper implementation.

Kyrrui:
I also propose an addition to this: allow the user to save a backup code associated with their device private key onto their file system, or would this confuse the user too much?

alexvandesande:
Interesting, didn't knew such options existed, sounds good. To reiterate: I'm against having the user download their local private key. That's what the backup codes are for.

moodysalem:
Local storage isn’t secure. For example, if one application using the SDK has a cross site scripting vulnerability, it can read local storage of the host and steal funds. Or, if the DNS is hacked, that can also expose local storage to attackers.

With my understanding of how universal login works, exposing the private key of a single app can expose a user’s funds in the proxy contract.

Instead, I would recommend moving universal login to a browser extension, and injecting an ethereum provider. This is an extra step that was deliberately avoided, but it’s the only way to persist a user’s authentication on a single website without exposing it to the site JavaScript and has the benefit that you authenticate the browser client instead of each individual app.

Heard about this issue in other apps in devcon4 talk by @tayvano

Kyrrui:
@moodysalem Are localDB and webSQL vulnerable as well? The library I use tries those two storage methods before it ever touches localstorage.

moodysalem:
Yes, and actually the problem isn’t just attackers targeting the website but also the developer of the website. E.g. if I’m malicious I might create a nice looking wallet and get users to log in to see their funds which gives me access to a private key to their proxy contract-even though my app may function perfectly well I now have access to a private key to do what I please.

I think that the storage and handling of the private key can’t happen in the same security context of the DAPP JavaScript. It puts too much trust in the DAPP developer to not be malicious nor hacked by a malicious user

[...]

PR #103, opened 2018-10-10
[WIP] [#93] Persistent storage & simple device display

Kyrrui:
I created a storage service which keeps you logged in even after refreshing or closing the browser. In order to easily clear your cache, I also implemented a simple version of the device list. To be added in later PR:
1. Testing for storageService
2. Listing of backup codes on device list
3. Logging out a device that was removed

### DIFF
--- a/universal-login-example/src/components/ContentContainer.js
+++ b/universal-login-example/src/components/ContentContainer.js
@@ -22,6 +22,7 @@ class ContentContainer extends Component {
   componentDidMount() {
     const {emitter} = this.props.services;
     this.subscription = emitter.addListener('setView', this.setView.bind(this));
+    this.loadIdentity();
   }
 
   componentWillUnmount() {
@@ -31,6 +32,15 @@ class ContentContainer extends Component {
   setView(view) {
     this.setState({view});
     window.scrollTo(0, 0);
+  }
+
+  async loadIdentity() {
+    const {emitter, storageService, identityService} = this.props.services;
+    let identity = await storageService.checkForIdentity();
+    if (identity) {
+      identityService.identity = identity;
+      this.setView('MainScreen');
+    }
   }
 
   render() {
@@ -43,7 +53,7 @@ class ContentContainer extends Component {
     } else if (this.state.view === 'MainScreen') {
       return <MainScreen services={this.props.services}/>;
     } else if (this.state.view === 'Account') {
-      return (<Account identityService={this.props.services.identityService}/>);
+      return (<Account services = {this.props.services}/>);
     } else if (this.state.view === 'ApproveConnection') {
       return <ApproveConnection services={this.props.services}/>;
     } else if (this.state.view === 'PendingAuthorizations') {
--- a/universal-login-example/src/components/ContentContainer.js
+++ b/universal-login-example/src/components/ContentContainer.js
@@ -35,7 +35,7 @@ class ContentContainer extends Component {
   }
 
   async loadIdentity() {
-    const {emitter, storageService, identityService} = this.props.services;
+    const {storageService, identityService} = this.props.services;
     let identity = await storageService.checkForIdentity();
     if (identity) {
       identityService.identity = identity;
--- a/universal-login-example/src/services/StorageService.js
+++ b/universal-login-example/src/services/StorageService.js
@@ -5,17 +5,17 @@ const DEVICES_KEY = 'universal-login-devices';
 
 class StorageService {
   checkForIdentity() {
-    return localforage.getItem(IDENTITY_KEY).then(function(value) {
+    return localforage.getItem(IDENTITY_KEY).then(function (value) {
       return value;
-    }).catch(function(err) {
+    }).catch(function() {
       return false;
     });
   }
 
   storeIdentity(identity) {
-    return localforage.setItem(IDENTITY_KEY, identity).then(function (value) {
+    return localforage.setItem(IDENTITY_KEY, identity).then(function () {
       return true;
-    }).catch(function(err) {
+    }).catch(function() {
       return false;
     });
   }
@@ -23,7 +23,7 @@ class StorageService {
   getDevices() {
     return localforage.getItem(DEVICES_KEY).then(function (value) {
       return value;
-    }).catch(function(err) {
+    }).catch(function() {
       return false;
     });
   }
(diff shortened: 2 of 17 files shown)


alexvandesande:
Very good start!

I noticed that if you make a connection request, that key is lost after a refresh, so that if you request in one device, accept it in another and go back to the original device after a refresh, that account will have been lost. We should probably store the keys (and state) as soon as you make a request too.

Kyrrui:
> Very good start!
> 
> I noticed that if you make a connection request, that key is lost after a refresh, so that if you request in one device, accept it in another and go back to the original device after a refresh, that account will have been lost. We should probably store the keys (and state) as soon as you make a request too.

I agree, to make this possible we will need to add a function to the SDK where you can ask for the management keys of a contracts, then when the user hits the login page, I do a check for identity in localstorage AND I see if the device in localstorage is a management key using that api. Then I can store the identity as soon as the device makes a request, otherwise the user will just be logged in (although unable to do anything) on a refresh.

alexvandesande:
Well you need to have that list anyway, in order to make the "devices accordion" work. It can be read directly from the contract, but you're right there should probably be a simple SDK call

alexvandesande:
> the user will just be logged in (although unable to do anything) on a refresh.

This behavior is actually what happens when you delete your own key. You are visibly logged in, but can't do anything. You should maybe refresh the page after removing a key?

Kyrrui:
> > the user will just be logged in (although unable to do anything) on a refresh.
> 
> This behavior is actually what happens when you delete your own key. You are visibly logged in, but can't do anything. You should maybe refresh the page after removing a key?

Hmm, that's odd - you should be redirected to the login screen when you delete your own key.

universal-login-example/src/views/DevicesModalView.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
```
alexvandesande: I feel we are overusing this modal. This whole code can be perfectly replaced by a "confirm()" javascript call, which is not only less code but more native. The current code doesn't look well on mobile.

screen shot 2018-10-10 at 2 01 52 pm

Kyrrui: Could you expand on this?

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
+import localforage from 'localforage';
+
+const IDENTITY_KEY = 'universal-login-identity';
+const DEVICES_KEY = 'universal-login-devices';
+
+class StorageService {
+  checkForIdentity() {
+    return localforage.getItem(IDENTITY_KEY).then(function (value) {
```
marekkirejczyk: Why not async/await?

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+  getDevices() {
+    return localforage.getItem(DEVICES_KEY).then(function (value) {
+      return value;
+    }).catch(function() {
+      return false;
+    });
+  }
+
+  async clearStorage() {
+    localforage.clear();
```
marekkirejczyk: Does clear returns promise? if so it should be:
`return localforage.clear();`
which in this case is the same as:
`await localforage.clear();`

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+    }).catch(function() {
+      return false;
+    });
+  }
+
+  getDevices() {
+    return localforage.getItem(DEVICES_KEY).then(function (value) {
+      return value;
+    }).catch(function() {
+      return false;
```
marekkirejczyk: Should return an empty array. Functions should always return the same type.
If there are items - an array of items if no items - empty array.

Same is true for other functions. Should return null, undefined or throw an exception.

Returning different types makes it harder to use, test and destroys JIT compiler optimisations.

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+        this.clearStorage();
+        return true;
+      }
+      // Find device in storage json object and delete
+      for (var i = 0; i < devices.length; i++) {
+        if (devices[i].publicKey == publicKey){
+          devices.splice(i, 1);
+          break;
+        }
+      }
```
marekkirejczyk: Shorter version: `devices.filter( (device) => device.publicKey !== publicKey);`

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+    }).catch(function() {
+      return false;
+    });
+  }
+
+  async removeDevice(publicKey) {
+    let devices = await this.getDevices();
+    if (devices) {
+      // If device removed is this device, clear identity
+      if (devices[0].publicKey == publicKey) {
```
marekkirejczyk: always use `===`

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+  async removeDevice(publicKey) {
+    let devices = await this.getDevices();
+    if (devices) {
+      // If device removed is this device, clear identity
+      if (devices[0].publicKey == publicKey) { 
+        this.clearStorage();
+        return true;
+      }
+      // Find device in storage json object and delete
+      for (var i = 0; i < devices.length; i++) {
```
marekkirejczyk: `for (const device of devices)` but see below

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,72 @@
...
+
+  async removeDevice(publicKey) {
+    let devices = await this.getDevices();
+    if (devices) {
+      // If device removed is this device, clear identity
+      if (devices[0].publicKey == publicKey) { 
+        this.clearStorage();
+        return true;
+      }
+      // Find device in storage json object and delete
```
marekkirejczyk: Using comments is often a bad small -> rather write your code in a shorter/more readable manner or extract a smaller method with an informative name

universal-login-example/src/components/ContentContainer.js
```diff
@@ -33,6 +34,15 @@ class ContentContainer extends Component {
...
   }
 
+  async loadIdentity() {
+    const {storageService, identityService} = this.props.services;
+    let identity = await storageService.checkForIdentity();
+    if (identity) {
+      identityService.identity = identity;
+      this.setView('MainScreen');
+    }
+  }
```
marekkirejczyk: There should be no logic in Reacgt components whatsoever. All the logic should be in services.
In particular, IdentityService might have a StorageService as a dependency (passed in constructor during build time).
All persistence should be implemented in StorageService and covered with roundtrip tests. 
In the top of that, we have IdentityService which implements logic like getting identity from storage or from sdk and updating state based on those.

Kyrrui:
@marekkirejczyk Cheers! I'm relatively new to React, so it's super helpful to learn things such as keeping business logic in the Services. I appreciate you bearing with some of my code hiccups along the way.

marekkirejczyk:
I am looking forward to merging this PR. @Kyrrui Let me know if need any help :)

Kyrrui:
Moving work to #114

PR #114, opened 2018-10-15
Adding localforage and StorageService

Kyrrui:
This is a cleaned up version of #103 - this does not contain device management, only identity loading and storage in the browser. I plan on writing tests for this, but first I need to create a new [custom localforage driver](https://github.com/localForage/localForage/wiki/Custom-Drivers)

### DIFF
--- a/universal-login-example/src/components/ContentContainer.js
+++ b/universal-login-example/src/components/ContentContainer.js
@@ -18,7 +18,6 @@ class ContentContainer extends Component {
       view: 'Login',
       viewParameters: {}
     };
-    this.loadIdentity();
   }
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


Kyrrui:
I decided to take out device management from this PR, because there is another bounty being worked on that would obsolete this method of device management (3box storage)

universal-login-example/src/components/ContentContainer.js
```diff
@@ -17,6 +17,7 @@ class ContentContainer extends Component {
     this.state = {
       view: 'Login'
     };
+    this.loadIdentity();
```
marekkirejczyk: I propose to move it to `Services.start`

universal-login-example/src/services/StorageService.js
```diff
@@ -0,0 +1,35 @@
...
+    return await localforage.setItem(IDENTITY_KEY, identity, async (err) => {
+      if (err) {
+        return false;
+      }
+      return true;
+    });
+  }
+
+  async clearStorage() {
+    return await localforage.clear();
```
marekkirejczyk: Shorter version:  
```
async clearStorage() {
    return localforage.clear();
}
```
You can skip await if you return promise form async function.

Kyrrui:
@marekkirejczyk Not sure what's up with the CI, but can't merge :/

PR #115, opened 2018-10-15
Small changes  to greetings and backup views

alexvandesande:
This PR adds a confirmation warning before the person exists the backup screen without saving it. Also removes the backup modal and instead replaces it by the greetings page

Future todo: while in this page, the app should provide a beforeUnload warning. I tried adding one, but it was unreliable, I think that while testing the browser prevents an app from showing one too many times. Also, the greetings view should reflect the reality of the user, stating if they have a saved backup or not.


alexvandesande:
I also added a temporary "greetMode" which will be able to show multiple checkmarks and I'm showing the greetings page more often.

I haven't figured out a way to change the different greet modes, so right now I'm just using this to hard code the change for some screen recordings for the demos

screen shot 2018-10-16 at 12 58 51 pm

PR #119, opened 2018-10-21
[WIP] General authorization

### LINKED ISSUE #88: Having a generic authorization contract for each key
Currently when you approve a new device to use your app, it is by default a "management key" which means that it has unlimited powers to do anything they want to the contract. The only limits on keys are given by the `purpose` and the `requiredApprovals` number set by ERC725 standard. This is insufficient. 

Consider the case in which you might want to add a phone that is able to add and approve new devices on the contract, but you want to set limits on how many new keys they can add in one day to prevent abuse. Or maybe you want to be able to add access to a very limited scope, maybe a key that can only interact with a single contract. To do that we need to either replace or improve the `key purpose` concept with an "authorization contract" that authorizes a key to do things according to a given set of rules.

We want to keep the types of authorization contracts open, so for the purpose of this issue we would simplify with 2 example authorization contracts:

* a contract that limits which contracts that key can interact with. It can be set to no contracts, only one specific contract, or any contract 
* a contract that limits the amount of "addKey" operations that each key can do, on the identity contract, per day. By default the logic would be: after adding N new keys, no new "addKeys" operations are accepted for N days

A proposed layout for how this would look for the user would be similar to the authorization screens of android or facebook apps, to be added on the "pendingAuthorization" screen:

screen shot 2018-10-04 at 10 57 14 am

shine2lay:
Goal
---
Add the general authorization described in : closes #88 

Changes
---
TODO: list all the changes in the PR

### DIFF
--- a/universal-login-contracts/contracts/SignedApprovalScheme.sol
+++ b/universal-login-contracts/contracts/SignedApprovalScheme.sol
@@ -87,10 +87,7 @@ contract SignedApprovalScheme is KeyHolder {
         require(executions[_id].to != address(this) || keyHasPurpose(signer, MANAGEMENT_KEY), "Management key required for actions on identity");
 
         executions[_id].approvals.push(bytes32(msg.sender));
-        if (executions[_id].approvals.length == requiredApprovals) {
-            return doExecute(_id);
-        }
-        return false;
+        return attemptExecution(_id);
     }
 
     function addSignedExecution(address _to, uint256 _value, bytes _data, uint256 _nonce, address _gasToken, uint _gasPrice, uint _gasLimit) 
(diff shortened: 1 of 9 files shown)


marekkirejczyk:
Hi @shine2lay!

Since the plan for authorization is not yet final I propose we treat it is a test/research implementation and move it to [proposals](https://github.com/UniversalLogin/Proposals) repository.

shine2lay:
@marekkirejczyk 
I think there are two parts to the authorizations,
1. The interface of authorization Contract
2. ways to handle complex multi-authorizations checking in SignedApprovalScheme.sol

I think the interface of the authorization contract is somewhat final (unless this implementation brings some issues we haven't considered to light).
My current work so far is to make sure that that interface is working and we can make one authorization check happen. Once this is working, adding multiple checks will work for sure no matter how we complex we decide to implement it.

I think the 2nd part is going to very complex and it may not even be necessary at the moment. I think it is something that will evolve overtime as requirements and use cases becomes clearer. We may even end up having different version of wallet contract that serves different purposes.

marekkirejczyk:
Would you like to make a call early next week to sync?

shine2lay:
@marekkirejczyk I can only do Thursday or Friday next week.

marekkirejczyk:
Hi, shot me an email on marek (at) ethwroks . io and let's start moving!

PR #137, opened 2018-11-07
Counterfactual instantiation of identity contract upon first execution

### LINKED ISSUE #90: Counterfactual deployment of contract address
One of the main drawbacks of the contract as your account is that it creates a fixed cost for user acquisition. This can be improved by "counterfactual deployment" or, a method in which you can deterministically sign a transaction in which you are certain of what will be your final address, in a way that it can be deployed by anyone later. It has the following advantages:

1. The user knows their address immediately and doesn't have to wait for a transaction to be mined
2. The same address can be generated on all networks, including private or test networks, so the app can onboard users on a cheaper network (or a side chain) and then migrate them to the main chain only when required
3. The address can refund the deployer and therefore only deploy once the address has funds to pay back
4. almost as a side effect, it also allows the user to select their icon and address from multiple generated ones.

### How would this affect this app

The user interface would be changed very little by this, as it's mostly something that happens under the hood. One interesting thing that could be done would be to add one screen to the "create account" view in which allowed the user to pick an icon from a list of 16 generated ones.

Perseverance:
Resolves #90. It is a merge from the fork. @JustynaBroniszewska @marekkirejczyk please check the TokenGrantingRelayer. I believe you've removed it but it was present in my repo and not sure how to migrate my code to the new changes

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,2 +1,4 @@
 {
+	"typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
+	"javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false
 }
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Before you issue pull request:
 You can run lint via:
 
 ```sh
-./script/lint.sh
+./scripts/lint.sh
 ```
 
 You can run tests:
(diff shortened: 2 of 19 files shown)


marekkirejczyk:
Hi @Perseverance,

I took a closer look yesterday at the PR, it will be hard to merge it all at once (failing tests, merge conflicts, etc)
I propose we split that into 3 smaller PRs:
- smaller updates (readme, eslint, etc)
- new classes with tests (CounterfactualIdentityService, CounterfactualTransactionsService, etc)
- and integration with the application (IdentityService, Relayer, etc)

TokenGrantingRelayer was moved to src/relayer. I suspect there is no need to modify it in anyway in your PR now.

Marek

Issue #99, opened 2018-10-10
Creating a new login for a domain that doesn't exist should fail nicely


abcoathup:
If you create a new login for a domain that doesn't exist, then the app keeps showing Creating ID page.

1. In the login, type bob.notexists.eth
2. Press the create button to create login bob.notexists.eth
3. App shows Creating ID page

Ideally creating a new login for a domain that doesn't exist should fail nicely with some sort of message that login notexists.eth doesn't exist or couldn't be found.

marekkirejczyk:
It is in our internal backlog

canepat:
@marekkirejczyk I would like to contribute on this issue if no one is already working on it.

As a start I would post a brief analysis with my comprehension of the problem and the proposed solution for review

marekkirejczyk:
@canepat Looking forward to your PR. Let me know if you would like to schedule a short call to discuss it.

marekkirejczyk:
Fixed by [143](https://github.com/UniversalLogin/UniversalLoginSDK/pull/143)

PR #143, opened 2018-11-16
Invalid ENS name error handling

canepat:
This PR addresses #99 

I've made the following working assumptions:
- client side: a simple dedicated Failure view is provided (purely additive)
- server side: the application errors are strings
- server side: the same error handling mechanism also for executeSigned API
- server side: one generic HTTP status 500 is fine for reporting errors

Possible improvements:
- using error codes shared between server and client instead of error strings (for i18n)
- using specific exception instead of generic Error (for additional info, not sure about advantages here)


marekkirejczyk:
LGTM, excellent job 👍 👍 👍

PR #145, opened 2018-11-19
3Box Integration

ChrisCates:
For https://github.com/UniversalLogin/UniversalLoginSDK/issues/102

I've noticed some issues for 3Box, please review:

- https://github.com/uport-project/3box-js/issues/184

- https://github.com/SilentCicero/ipfs-mini/issues/10

Notably, there are no options to set an optional host for ipfs-mini for threebox itself. I believe this is something that should be available when running the constructor method for 3box itself. I am open to writing pull requests to both ipfs-mini and 3box-js for this to actually work.

I've also updated the UI here:

### DIFF
--- a/universal-login-example/src/services/Threebox.js
+++ b/universal-login-example/src/services/Threebox.js
@@ -1,0 +1,22 @@
+import Box from '3box';
+
+class ThreeBox {
+  constructor(provider) {
+    console.log(Box);
+    this.provider = provider;
+    this.enabled = false;
+    this.box = null;
+    this.addresses = [];
+  }
+
+  async enable() {
+    this.addresses = await window.ethereum.enable();
+    this.enabled = true;
+  }
+
+  async open() {
+    this.box = await Box.openBox(this.addresses[0], window.ethereum);
+  }
+}
+
+export default ThreeBox;
--- a/universal-login-example/package.json
+++ b/universal-login-example/package.json
@@ -4,6 +4,8 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
+    "3box": "^1.0.5",
+    "classnames": "^2.2.6",
     "components": "^0.1.0",
     "date-fns": "^1.29.0",
     "detect-browser": "^3.0.1",
--- a/universal-login-example/scripts/dev-start.js
+++ b/universal-login-example/scripts/dev-start.js
@@ -105,7 +105,7 @@ class Deployer {
   }
 
   startRelayer() {
-    this.relayer = new TokenGrantingRelayer({...this.config, privateKey: this.deployerPrivateKey, tokenContractAddress: this.tokenContract.address}, this.provider);    
+    this.relayer = new TokenGrantingRelayer({...this.config, privateKey: this.deployerPrivateKey, tokenContractAddress: this.tokenContract.address}, this.provider);
     this.env.RELAYER_URL = `http://${args.httpAddress}:${this.config.port}`;
     this.relayer.start();
   }
--- a/universal-login-example/src/components/MainScreen.js
+++ b/universal-login-example/src/components/MainScreen.js
@@ -62,7 +62,7 @@ class MainScreen extends Component {
           <RequestsBadge setView={this.setView.bind(this)} services={this.props.services}/>
           <AccountLink setView={this.setView.bind(this)} />
         </HeaderView>
-        <MainScreenView clicksLeft={this.state.clicksLeft} events={this.state.events} onClickerClick={this.onClickerClick.bind(this)} lastClick={this.state.lastClick} />
+        <MainScreenView clicksLeft={this.state.clicksLeft} events={this.state.events} onClickerClick={this.onClickerClick.bind(this)} lastClick={this.state.lastClick} services={this.props.services} />
       </div>
     );
   }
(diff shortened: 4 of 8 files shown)


ChrisCates:
@alexvandesande, due to how 3box works... Would we be using Mascara instead?

Yes, the buttons are tbd.

If we end up using mascara, it probably won't need an "enable Metamask" button.

universal-login-example/src/services/Threebox.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor(provider) {
+    console.log(Box);
+    this.provider = provider;
+    this.enabled = false;
+    this.box = null;
+    this.addresses = [];
+  }
+
+  async enable() {
+    this.addresses = await window.ethereum.enable();
```
alexvandesande: This is where you probably need to change. There is no concept of window.ethereum here–but then maybe it should have.

@marekkirejczyk what if instead of having to change 3box here, don't we change the SDK to provide an ethereum object by default and it will pipe through all the transaction requests to either a meta transaction for the identity or a simple signature by one of the local keys.

ChrisCates: @alexvandesande, we should integrate this instead: https://github.com/MetaMask/mascara

alexvandesande: I'm sorry, Chris, there seems to be a misunderstanding on what Universal Logins is supposed to be. We are not integrating Mascara. Universal logins sorts of makes Mascara (or Metamask, Mist, etc) redundant (or only as an optional add-on for extra security).

The app already has a local private key. We should use that for signing messages instead.

ChrisCates: This is my bad.
I got confused from the 3box-js documentation. I assumed that metamask was necessary. I was aware of what the repository was trying to do... I was confused on how to make the implementation without Metamask.

**From the 3box-js documentation:**

> ### Get, set, and remove data
> To get or modify data in a user's 3Box, first open their 3Box by calling the openBox method. This method prompts the user to authenticate your dapp and returns a promise with a threeBox instance. You can only set, get, and remove data of users that are currently interacting with your dapp. Below ethereumProvider refers to the object that you would get from web3.currentProvider, or window.ethereum.

When you prompt the `.openBox()` method. It opens a dialog in Metamask to allow the dApp to work. In this case, we just need to pop in the provider for the existing web3 instance.

alexvandesande: Chris, you are correct, that's how you use 3Box with a ethereum enabled browser. Your post convinced me that, instead of changing the 3box code, we should instead thrive to have a `web3.currentProvider` or `window.ethereum` objects on our app that simulates what the tools expect from metamask

universal-login-example/src/services/Threebox.js
```diff
@@ -0,0 +1,22 @@
+import Box from '3box';
+
+class ThreeBox {
+  constructor(provider) {
+    console.log(Box);
```
marekkirejczyk: There should be no console.log in production code

ChrisCates: @marekkirejczyk, was intended for debugging purposes. But, thanks for the heads up.
The code there right now is for reviewing purposes (as I was confused about the best method of implementation at the time).

alexvandesande:
Had a chat with @marekkirejczyk about the possibility of having universal logins always add by default an ethereum object. He thinks adding a global object could lead to many undesired collateral effects. Another possibility would be, in this particular case, create an ethereum provider specifically for threeBox.js that would send any personal_sign transaction towards the same functions at ether.js

ChrisCates:
Hey @alexvandesande, sorry this went in limbo for a few days.
It seems they updated the API the day after we had the conversation in regards to the problems with IPFS on Infura.

Here's the updated docs: https://github.com/uport-project/3box-js#Box.openBox

We also will need a service for a 3Box Pinning Server and Name Server:

- https://github.com/uport-project/3box-pinning-server
- https://github.com/uport-project/3box-address-server

If this makes sense, I'm going to update it with the Provider passed into the SDK.
I'm going to scrap this PR for a new one.

universal-login-example/src/services/Threebox.js
ChrisCates: @alexvandesande, I'm working on something now that ideally will be able to work Metamask or Mascara. I will give you an update if there are problems. Or a PR for review if no issues.

oed:
@ChrisCates @alexvandesande There is another option, that is not documented unfortunately, that allows you to create a 3box without having an ethereum account:

Internally 3box uses an indentity construct called "muport". You can create a muport identity without an ethereum account from a mnemonic. This muport can then be passed to the 3Box constructor. See the code below:
```js
const mnemonic = bip39.generateMnemonic() // you can also derive the mnemonic from a base entropy, see the bip39 docs
const id = await MuPort.newIdentity({ name: 'name123' }, null, {
  mnemonic
})
const box = new Box(id)
await box._load() // this function has to be called in order for 3box to work correctly.
box._linkProfile = () => {} // this disables any further calls to the ethereum client
```
Note that this 3box will be independent and not be linked to any ethereum address.

Hope this is helpful in what you are trying to do. Let me know if you have any questions!

PR #172, opened 2018-12-06
Make UI changes to make it better on async environments

alexvandesande:
This makes a few changes to the code so that the experience is better on a real network. Things like:

* Creates more intermediary steps to show creation of account
* Adds more loading css classes that show things that are loading
* Reflects some state changes immediately and then waits for the answer from the network

Work in progress. Things to do:

- [ ] Need a way to show pending transactions on the history window
- [x] Need to refactor the greetingView so it doesn't replicate code from CreatingID view
- [ ] Refactor the connection hover so that it appears more fast, even without knowing how to connect to it

### DIFF
--- a/universal-login-example/src/components/Greeting.js
+++ b/universal-login-example/src/components/Greeting.js
@@ -13,6 +13,7 @@ class Greeting extends Component {
   }
 
   async componentDidMount() {
+    console.log(this.state);
     const keyHolderAddress = this.props.identityService.identity.address;
     const {greetMode} = this.props.viewParameters;
     const status = await this.props.greetingService.getStatus(keyHolderAddress, greetMode);
--- a/universal-login-example/src/components/Greeting.js
+++ b/universal-login-example/src/components/Greeting.js
@@ -13,7 +13,6 @@ class Greeting extends Component {
   }
 
   async componentDidMount() {
-    console.log(this.state);
     const keyHolderAddress = this.props.identityService.identity.address;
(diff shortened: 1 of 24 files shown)


alexvandesande:
@marekkirejczyk I'd like to request help on something that you'll do much faster than I can figure it out. I think a big pain point on the experience is at the moment when you first encounter the app, that it takes some time to load the hover box and often shows an older name on there. I think the reason is that it only shows the results when it gets a return from checking all identity names, and I think it should be different. 

I suggest that you should show all names immediately, even if you don't know their connection state, and only then change that when you have that information. Just show the names with a disabled or blank button besides them.

Also, often the suggested username is different than the one you received often happens when you are typing fast, sometimes an intermediary name is the last to answer). I suggest either:

* Add a small timer on the onchange, so that it only actually queries once you haven't typed anything in x miliseconds
* Do not update the state if the name you are receiving does not correspond with the one on the input field
* At each key stroke, cancel the previous requests

alexvandesande:
Made a few more changes to make the UI simpler and more consistent. Besides the described changes to the connection hover, the only thing I'd add is that I find particularly weird that whenever the page loads, it takes almost exactly 30 seconds for it to receive new events. Is it waiting for a new block?

marekkirejczyk:
Updated and refactored code a bit. Will merge it now.
Two more topics to be covered in speparate PRs:
- pending history
- refactor the connection hover

PR #173, opened 2018-12-07
main screen improvements

JustynaBroniszewska:


universal-login-example/src/services/ClickerService.js
```diff
@@ -46,31 +49,61 @@ class ClickerService {
     return await this.ensService.getEnsName(address);
   }
 
-  async getEventsFromLogs(events) {
+  async getPressersFromLogs(events) {
```
marekkirejczyk: map instead of for?

marekkirejczyk:
Will continue on `async-ensnames` branch.

PR #175, opened 2018-12-13
Navigation of login hover view using keyboard

canepat:
This PR adds navigation of the login suggestion view using tab/up/down keys.

The code for moving selection up and down is not as clean as I would have liked, so any hint or cleaner solution is welcome.

I've also added a hard-coded limit of `24` characters for the username input textbox just to avoid hidden ENS subdomains on suggestion view when entered text is longer. Sadly Chrome on mobile still has a bug on this and allows inserting more than `maxLength` characters (even if it then truncates them if loses and regains focus).


canepat:
I've added `arrow up/down` key handling for moving focus to/from the input text box as suggested by @alexvandesande and the resulting user experience seems much better now.

Moreover I've also removed an unused property in one UI component and corrected a leak.

I've run the following commands

- `./script/lint.sh`
- `./script/test.sh`
- `./script/clean.sh`
- `yarn`
- `./script/travis.sh`

successfully on local repository and I still can't figure out why the Travis CI build failed

canepat:
It seems that the test `subscribe: should emit AddKey on construction` in `BlockchainObserver.js` may sometimes fail, it happened also on my local repository. I don't see any relationship with this PR

alexvandesande:
Thanks for the PR, indeed the experience is much better now. 

However I don't understand why these changes are so more complex than needed. You make changes on LegacyENS and add a function to save variables to a file, which I could not understand how are related to the keyboard support. Could you expand on what you're trying to do there?

canepat:
I guess the changes you're referring to are coming from commit 2adb75b572c0e4210f8b4915edca35ad3a31da94 which I've got when fetching and merging the changes occurred on `master` after my first commit in this PR. Maybe I've messed with rebase here, sorry if so.

My last changes after your suggestions are in commits d2e3c041d300efc4105e29770b8d08b8c20150e4 and ba57327a52d5e807bb8b38ad8398766bc5dcc4d7

marekkirejczyk:
Manually merged in PR [#189](https://github.com/UniversalLogin/UniversalLoginSDK/pull/189)

marekkirejczyk:
Great work! Thank you @canepat for great pair programming session :)

PR #195, opened 2019-01-14
Change git/github repository URLs to new repository

yaram:
I don't know if there's a reason for them being the old URL, but it seemed like a small easy change.


marekkirejczyk:
Good point. Thank you! 👍

PR #200, opened 2019-01-17
addKey can also be used to add to the identity all of the avaible types of keys

Giulio2002:
While studying the functioning of action keys and management keys i noticed that sdk.addKey and sdk.addKeys, can only be used to add management keys to the identity, therefore to study action keys i modified sdk.js so that addKey now needs a new argument(keyType) that is the type of the new key( Management, Action, Encryption) which is set by default to MANAGEMENT_KEY. modifying that variable we can choose if we wants to add a Management, Action or Encryption key. i decided to create this pr just because i thought it might be useful.

### DIFF
--- a/universal-login-sdk/lib/sdk.js
+++ b/universal-login-sdk/lib/sdk.js
@@ -34,9 +34,9 @@ class EthereumIdentitySDK {
     throw new Error(`${responseJson.error}`);
   }
 
-  async addKey(to, publicKey, privateKey, transactionDetails) {
+  async addKey(to, publicKey, privateKey, transactionDetails, keyPurpose = MANAGEMENT_KEY) {
     const key = addressToBytes32(publicKey);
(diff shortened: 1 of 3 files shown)


JustynaBroniszewska:
Great work! 👍 We would love to merge it! 😄 

But first I would propose some improvements: 
1. MANAGEMENT, ACTION etc. are key purposes, not key types. Please, change keyType to keyPurpose.
2. Do we know that it works properly? Maybe you could write two tests to check it? One for add one key, and the second one for adding multiple keys.

Giulio2002:
i made the changes that you suggested and i also modified some of the already present tests(just the ones that tests the addKey method), because some of them don't check if the contract's state have been modified or not.

JustynaBroniszewska:
Awesome! Love it! 👍 😍

PR #236, opened 2019-02-06
bump ops version; update script name

marekkirejczyk:


marekkirejczyk:
Broken build due to dependencies. Reverting to last working version.

PR #255, opened 2019-02-08
Migrations after start in relayer

JustynaBroniszewska:


docs/source/relayer.rst
```diff
@@ -29,12 +29,6 @@ You can do it in your favorite database UI, or from `psql`:
     > create database universal_login_relayer_development;
     > \q
 
-Then you need to migrate database:
-
-  ::
-
-    yarn setup:db
-
```
marekkirejczyk: Do we remove tasks as well?

universal-login-example/src/relayer/startRelayer.js
```diff
@@ -6,4 +6,4 @@ const config = require('./config');
 const db = getKnex();
 
 const relayer = new TokenGrantingRelayer(config, db, '');
-relayer.database.migrate.latest().then(() => relayer.start());
```
marekkirejczyk: relayer.start().catch(error.console)

universal-login-relayer/lib/relayer.js
```diff
@@ -32,13 +31,9 @@ class Relayer {
   }
 
   async start() {
-    if (await checkIfAllMigrated(this.database)) {
```
marekkirejczyk: Delete `checkIfAllMigrated` if not used

PR #263, opened 2019-02-11
getKnex moved to relayer constructor

JustynaBroniszewska:


universal-login-relayer/lib/config/knexfile.js
```diff
@@ -9,6 +16,7 @@ module.exports = {
     },
     migrations: {
       tableName: 'knex_migrations',
+      directory: getMigrationPath(),
```
marekkirejczyk: Can we use a relative path here?

universal-login-relayer/lib/config/knexfile.js
```diff
@@ -19,7 +27,9 @@ module.exports = {
       user:     'postgres',
       password: 'postgres',
     },
-    migrations: {},
+    migrations: {
+      directory: getMigrationPath(),
```
marekkirejczyk: Can we use a relative path here?

universal-login-ops/src/index.js
```diff
@@ -0,0 +1,6 @@
+const startDevelopment = require('./dev/startDevelopment');
+const createEnv = require('./common/createEnv');
+const spawnProcess = require('./common/spawnProcess');
+const TokenGrantingRelayer = require('./dev/TokenGrantingRelayer');
```
marekkirejczyk: You can use `import` now :)

PR #272, opened 2019-02-13
Wallet: Services added

JustynaBroniszewska:


universal-login-wallet/src/services/Services.ts
```diff
@@ -0,0 +1,17 @@
...
+interface Config {
+  relayerUrl: string;
+  jsonRpcUrl: string;
+}
+
+export interface Services {
+  sdk: any;
+}
+
+const Services = (config: Config) => {
```
sz-piotr: This function being uppercase is misleading. If I were using it I would assume it's a class. You can change it to `createServices` **unless** it is the same in other projects - consistency is more important.

universal-login-wallet/src/services/Services.ts
```diff
@@ -0,0 +1,17 @@
+const UniversalLoginSDK = require('universal-login-sdk').default;
```
sz-piotr: `import UniversalLoginSDK from 'universal-login-sdk';`

universal-login-wallet/test/Services.ts
```diff
@@ -0,0 +1,10 @@
+import {expect} from 'chai';
+import createServices from '../src/services/Services';
+
+describe('Services', () => {
+  it('Should create services', () => {
+    const services = createServices({relayerUrl: 'http://relayer.com', jsonRpcUrl: 'http://mockprovider.com'});
+    expect(services).to.not.be.null;
+    expect(services.sdk).to.not.be.null;
+  });
+});
```
sz-piotr: Doesn't the linter require newlines?

JustynaBroniszewska: Now it does 😃

PR #279, opened 2019-02-14
[WIP] Added support for other databases (phase 1)

### LINKED ISSUE #278: Add support for other databases besides PostgreSQL
The database for the authorisation service should be customisable and not force postgreSQL.

santiagorp:
# Summary
Current relayer requires PostgreSQL to work. Using a different DB such as MongoDB it is not possible with the current code since the file `universal-login-relayer/lib/relayer.js` uses knex directly.

This solution wraps database functionality into a separate database module which is now passed to the relayer constructor. The module needs to have the following methods:

* addRequest
* getPendingAuthorisations
* removeRequest
* onStart
* onStop

which are the methods required by the `authorisationService` and the relayer itself in the start/stop methods.

This solution modifies the `knexUtils.js` module into the `postgreDB.js` module which contains the previous methods.

In my opinion all database related code should be moved out of the project `universal-login-relayer` and create database specific projects named `universal-login-postgresql-relayer` or `universal-login-mongodb-relayer` depending on the desired implementation.

With this initial set of changes the usage of mongoDB or any other DBMS is a trival task.

Fixes #278 

## Checklist


santiagorp:
I see thre are some fixes pending in order to have this merged into master. Please let me know if you agree with the idea before fixing them.

marekkirejczyk:
Hi @santiagorp,

Thank you for the PR and splitting into small chunks :)
Also sorry for the slow response - we have been occupied with EthDenver, and now we are preparing for EthCC.

Support for multiple databases is something we can discuss,  there is a number of considerations to take into account:
- `TokenGrantingRelayer`,`DevelopmentRelayer` and `Relayer` should be constructible from just a config file. 
- that means that it should be able to construct itself with alternative database base on configuration (database factory)
- Perhaps `Relayer` could have additional parameter overrides that can override database (and ignore/skip config)
- I am not sure if it makes sense to introduce PostgreDB and extract authorisationService.js, as AuthorisationService is already thin persistency layer, perhaps we should call it AuthorisationRepository to make it clear

As an alternative, we can say Postgres is default choice and if one wants to use another database, then one should inherit or create an alternative implementation. Relayer sole responsibility is construction/managing lifecycle. 

I am looking forward to your thoughts.

Marek

marekkirejczyk:
Any thoughts @santiagorp?

santiagorp:
Hello @marekkirejczyk,

> Any thoughts @santiagorp?

Thanks for the reminder!

> Perhaps Relayer could have additional parameter overrides that can override database (and ignore/skip config)

I like this idea since it gives the possibility of adding custom database support. However the fact of having an unncessary dependency due to the underlying default implementation is far from ideal. Thus having a way to **inject the database** (or any other implementation) is more appropriate.

The current code would work by using the current postgre/knex implementation and in the future further implementations could be added to the repository as well.

BUT, I see that this might enter in conflict with the another statement:

> TokenGrantingRelayer,DevelopmentRelayer and Relayer should be constructible from just a config file.

Why is this a requirement? Why the database implementation cannot be passed to their constructor as well as the config file? As soon as the current example and tests have implemented one (like the current postgre version), it should not be a problem.

marekkirejczyk:
> >TokenGrantingRelayer,DevelopmentRelayer and Relayer should be constructible from just a config file.

> Why is this a requirement? Why the database implementation cannot be passed to their constructor as well as the config file? As soon as the current example and tests have implemented one (like the current postgre version), it should not be a problem.

So that we can set-up relayer from command line, create custom ops scripts or configure deploy easily.

marekkirejczyk:
Seems stalled. Closing for now. Feel free to reopen if you would like to continue working on this.

PR #315, opened 2019-03-01
added multi-signatures support

Giulio2002:
# Summary
While i was building a personal project of mine, i built a contract that split a bytes variable in three bytes32 variables, because in ERC1077.sol the variable signatures(according to what is said in its and ERC1078 EIP description) is supposed to carry more than one signature i adapted my code to change the canExecute function in ERC1077.sol. now the canExecute split the variable signatures in signatures.length / 65  signatures and verify them. if there are two equal signature contained in signatures canExecute return false. i added test for this new feature and a function in calcuateMessageSignature.ts that helps us to concatenate the signatures and testing the multi-signatures support. i didn't add any documentation. let me know if you look for it.

## Checklist
- [X] Change is small and easy to review (please split big changes into multiple PRs)
- [X] Change is consistent with architecture
- [X] Change is consistent with test architecture
- [X] Change is consistent with naming conventions
- [X] New code is covered with tests
- [X] Tests related to old code are updated


universal-login-contracts/test/utils.js
```diff
@@ -9,11 +8,7 @@ const addressToBytes32 = (address) =>
 const waitForContractDeploy = async (providerOrWallet, contractJSON, tansactionHash, tick = 1000) => {
   const provider = providerOrWallet.provider ? providerOrWallet.provider : providerOrWallet;
   const abi = contractJSON.interface;
-  let receipt = await provider.getTransactionReceipt(tansactionHash);
-  while (!receipt) {
-    sleep(tick);
-    receipt = await provider.getTransactionReceipt(tansactionHash);
-  }
+  const receipt = await provider.waitForTransaction(tansactionHash);
```
marekkirejczyk: 👍

universal-login-contracts/lib/calculateMessageSignature.ts
```diff
@@ -19,8 +19,11 @@ export const calculateMessageHash = (msg: Message) => {
     [msg.from, msg.to, msg.value, dataHash, msg.nonce, msg.gasPrice, msg.gasToken, msg.gasLimit, msg.operationType]);
 };
 
-export const calculateMessageSignature = (privateKey: string, msg: Message) => {
+export const calculateMessageSignature = async (privateKey: string, msg: Message) => {
```
marekkirejczyk: 👍

universal-login-contracts/test/contracts/ERC1077.js
```diff
@@ -30,13 +30,13 @@ describe('ERC1077', async  () => {
   let invalidSignature;
   let relayerBalance;
   let relayerTokenBalance;
-
+  let privateActionKey;
   beforeEach(async () => {
-    ({provider, identity, privateKey, keyAsAddress, publicKey, mockToken, mockContract, wallet} = await loadFixture(basicIdentity));
+    ({provider, identity, privateKey, privateActionKey, keyAsAddress, publicKey, mockToken, mockContract, wallet} = await loadFixture(basicIdentity));
     msg = {...transferMessage, from: identity.address};
     signature = calculateMessageSignature(privateKey, msg);
```
marekkirejczyk: Shouldn't `await` be used here as well?

universal-login-contracts/test/contracts/ERC1077-gasCost.js
```diff
@@ -9,10 +9,10 @@ import {getExecutionArgs} from '../utils';
 chai.use(chaiAsPromised);
 chai.use(solidity);
 
-const callCost = 109500;
-const transferCost = 125000;
+const callCost = 129500;
+const transferCost = 225000;
```
marekkirejczyk: This is part of continues performance practice - a test that makes sure we don't increase the cost of (already pretty expensive)  transaction significantly. 
We should:
- consider optimisations to keep single signature cost at the same/similar level
- add a test for multisignature case

universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -105,6 +106,8 @@ contract ERC1077 is KeyHolder, IERC1077 {
         OperationType operationType,
         bytes memory signatures) public returns (bytes32)
     {
+        require(signatures.length != 0, "Invalid signatures");
+        require(signatures.length % 65 == 0, "Invalid signatures");
```
marekkirejczyk: 👍 we should add tests for those cases

universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -31,18 +31,19 @@ contract ERC1077 is KeyHolder, IERC1077 {
...
-            data,
-            nonce,
-            gasPrice,
-            gasToken,
-            gasLimit,
-            operationType,
-            signatures);
-        return keyExist(bytes32(uint256(signer)));
+        address[] memory signers = new address[](signatures.length.div(65));
+        for (uint i = 0;i < signatures.length.div(65); i++) {
```
marekkirejczyk: space missing after `;`

universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -31,18 +31,19 @@ contract ERC1077 is KeyHolder, IERC1077 {
...
-            value,
-            data,
-            nonce,
-            gasPrice,
-            gasToken,
-            gasLimit,
-            operationType,
-            signatures);
-        return keyExist(bytes32(uint256(signer)));
+        address[] memory signers = new address[](signatures.length.div(65));
```
marekkirejczyk: We should add check for number of signatures to match required signatures, so:
- add new contract variable `uint requiredSignatures = 1` 
- add a private setter: `function setRequiredSignatures(value) onlyManagementKeyOrThisContract {...}`
- add a check at the beginning of the function: `require(requiredSignatures == signatures.length.div(65))`

universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -31,18 +31,19 @@ contract ERC1077 is KeyHolder, IERC1077 {
...
+        for (uint i = 0;i < signatures.length.div(65); i++) {
+            bytes memory sig = new bytes(65);
+            for (uint j = 0;j < 65; j++) {
```
marekkirejczyk: This is pretty expensive, should be done using assembly, extracted to separate method and tested.
Example of such a function can be found [here](https://github.com/gnosis/safe-contracts/blob/a97c6fd24f79c0b159ddd25a10a2ebd3ea2ef926/contracts/common/SignatureDecoder.sol#L33) and usage [here](https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafe.sol#L147)

universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -128,4 +131,13 @@ contract ERC1077 is KeyHolder, IERC1077 {
             msg.sender.transfer(gasUsed.mul(gasPrice));
         }
     }
+
+    function isSignerValid(address signer, address[] memory signers) private view returns (bool) {
+        for (uint i = 0;i < signers.length; i++) {
```
marekkirejczyk: This is quite expensive, we can avoid O(N^2) comparisons by doing a little trick:
- always sort signers in the same order (in javascript as they are passed as signatures, requires a helper js function)
- require each signer to be different than previous
- require each signer to be >= then previous (you can compare addresses same way as integers in solidity)

universal-login-contracts/test/contracts/ERC1077.js
```diff
@@ -202,6 +202,59 @@ describe('ERC1077', async  () => {
     });
   });
 
+  describe('Multi-Signature call', async () => {
```
marekkirejczyk: We should split into two describe: `Multi-Signature call with 2 signatures` and `Multi-Signature call with 3 signatures`.

universal-login-contracts/test/contracts/ERC1077.js
```diff
@@ -202,6 +202,59 @@ describe('ERC1077', async  () => {
...
+          .withArgs(messageHash, 0, true);
+      });
+    });
+
+    describe('failed execution of call via multi-signature', async () => {
+      before(async () => {
+        msgToCall = {...callMessage, from: identity.address, to: mockContract.address};
+      });
+
+      it('should fail with a least an invalid signature', async () => {
```
marekkirejczyk: We should add a test for invalid length.

universal-login-contracts/test/fixtures/basicIdentity.js
```diff
@@ -1,17 +1,20 @@
...
-import {OPERATION_CALL} from '../../lib/consts';
+import {OPERATION_CALL, ACTION_KEY} from '../../lib/consts';
 import DEFAULT_PAYMENT_OPTIONS from '../../lib/defaultPaymentOptions';
 
 const {parseEther} = utils;
 const {gasPrice, gasLimit} = DEFAULT_PAYMENT_OPTIONS;
 
 export default async function basicIdentity(provider, [, , , , , , , , , wallet]) {
   const publicKey = addressToBytes32(wallet.address);
+  const actionWallet = Wallet.createRandom();
```
marekkirejczyk: 👍 A nice alternative way of creating keys. We should use that instead of getWallets, whenever we want a key without balance on it.

universal-login-contracts/contracts/ERC1077.sol
marekkirejczyk: Perhaps in next PR - that might be a lot of additional testing.

universal-login-contracts/test/utilsTests.ts
```diff
@@ -26,6 +27,13 @@ describe('Tools test', async () => {
     expect(utils.verifyMessage(message, signature)).to.eq(wallet.address);
   });
 
+  it('Should concatenate two bytes arrays', async () => {
```
marekkirejczyk: I would add tests (and implementation) for cases when `bytes1` or `bytes2` doesn't start with 0x or is not a hex. Throwing an exception I would consider proper behavior.

universal-login-contracts/contracts/ERC1077.sol
Giulio2002: so for the moment it's better to not create the requiredSignature in this PR. right?

universal-login-contracts/contracts/ERC1077.sol
Giulio2002: Thank you. that helps a lot.

Giulio2002: Solium doesn't like solidity inline assemby so i turned off "security/no-inline-assembly" in .soliumrc.json

Giulio2002:
i have done as you said. hope to hear you soon.

universal-login-contracts/lib/calculateMessageSignature.ts
```diff
@@ -19,8 +19,20 @@ export const calculateMessageHash = (msg: Message) => {
...
+export const concatenateSignatures = (signatures: string[]) => {
+  let concatenated = '';
+  // sort the signatures
+  signatures.forEach((signature) => {
+    if (signature.indexOf('0x') !== 0) {
+      throw `Invalid Signature: ${signature} needs prefix 0x`;
+    }
+    concatenated += signature.replace('0x', '');
+  });
+  return `0x${concatenated}`;
```
marekkirejczyk: I would propose following implementation: 
```js
const removePrefix = (prefix: string, value: string)  => {
  if (value.length !== 67) {
    throw `Invalid signature length: ${signature} should be 67`;
  }
  if (value.indexOf('0x') !== 0) {
    throw `Invalid Signature: ${signature} needs prefix 0x`;
  }
  return signature.slice(2);
}

const concatenateSignatures = (signatures: string[])  => {
  return `0x${signatures.map(removePrefix).join('')}`; 
}
```

Giulio2002:
Done. However the tests input variables became long

Giulio2002:
Now it should be good

marekkirejczyk:
@Giulio2002 just want to say this PR is high on my priority list, but I am now in Paris and it gets pretty intensive. I want to reserve a bit longer quality time to review. It is small PR, but it is an important part of core functionality, so I want to get it right. Will get there soon.

universal-login-contracts/.soliumrc.json
```diff
@@ -4,6 +4,7 @@
     "security"
   ],
   "rules": {
+    "security/no-inline-assembly": "off",
```
marekkirejczyk: Would be better to make it a local exception, with 
```
/* solium-disable */
...
/* solium-enable */
```

universal-login-contracts/test/fixtures/basicIdentity.js
```diff
@@ -1,15 +1,31 @@
...
+import {OPERATION_CALL, ACTION_KEY} from '../../lib/consts';
 import DEFAULT_PAYMENT_OPTIONS from '../../lib/defaultPaymentOptions';
 
 const {parseEther} = utils;
 const {gasPrice, gasLimit} = DEFAULT_PAYMENT_OPTIONS;
 
 export default async function basicIdentity(provider, [, , , , , , , , , wallet]) {
+  const actionWallet1 = Wallet.createRandom();
+  const actionWallet2 = Wallet.createRandom();
+  const sortedWallets = [actionWallet1, actionWallet2, wallet].sort((wallet1, wallet2) => {
```
marekkirejczyk: I would extract it to separate utility function `sortWallets`.

marekkirejczyk:
This is shaping really nicely. Added some additional minor comments.
I want to read it one more time before merging.
Also @JustynaBroniszewska, would you take a look at it?

Thanks for all the hard work so far :) ❤️❤️ ❤️ ❤️

Giulio2002:
i converted sortWallets into sortPrivateKeysByAddress, so that if we a privateKey array like in the sdk we don't have to convert it to a wallet array and the convert it back

Giulio2002:
I have also made possible the multi signature into the sdk, so that everyone can try it

universal-login-contracts/lib/calculateMessageSignature.ts
```diff
@@ -19,8 +19,45 @@ export const calculateMessageHash = (msg: Message) => {
...
-export const calculateMessageSignature = (privateKey: string, msg: Message) => {
+export const calculateMessageSignature = async (privateKey: string, msg: Message) => {
   const wallet = new Wallet(privateKey);
   const massageHash = calculateMessageHash(msg);
   return wallet.signMessage(utils.arrayify(massageHash));
 };
+
+
+export const calculateMessageSignatures = async (privateKeys: string[], msg: Message) => {
+  const signatures = privateKeys.map((value: string) =>
```
marekkirejczyk: I would improve naming:
`const signaturesPromises = privateKeys.map((key: string) => calculateMessageSignature(key, msg));`

universal-login-contracts/test/contracts/ERC1077-gasCost.js
```diff
@@ -12,7 +12,7 @@ chai.use(solidity);
 const callCost = 90000;
 const transferCost = 105000;
 
-const overrideOptions = {gasLimit: 120000};
+const overrideOptions = {gasLimit: 220000};
```
marekkirejczyk: I think this change is now unnecessary.

universal-login-wallet/package.json
```diff
@@ -21,7 +21,7 @@
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
     "universal-login-commons": "^1.0.0",
-    "universal-login-sdk": "^0.1.11"
+    "universal-login-sdk": "^0.1.9"
```
marekkirejczyk: Should be 11 or 12.

universal-login-sdk/lib/sdk.js
```diff
@@ -34,7 +35,7 @@ class EthereumIdentitySDK {
     throw new Error(`${responseJson.error}`);
   }
 
-  async addKey(to, publicKey, privateKey, transactionDetails, keyPurpose = MANAGEMENT_KEY) {
+  async addKey(to, publicKey, privateKeys, transactionDetails, keyPurpose = MANAGEMENT_KEY) {
```
marekkirejczyk: This is not a good idea. That means that one client/device has access to all need keys, which contradicts the purpose of multisig. 

The way to implement it on SDK level is:
- one device requests an execution
- other devices get a notification
- other devices can confirm (by signing the same data and sending the signature to relayer)
- relayer gathers signatures and triggers execution when complete

Let's revert those changes now and go back to them in the next PR. 
We can do a call to discuss it.

I will use tag #mutipleKeys, to mark related changes.

universal-login-sdk/lib/sdk.js
```diff
@@ -43,10 +44,10 @@ class EthereumIdentitySDK {
       from: to,
       data,
     };
-    return this.execute(message, privateKey);
+    return this.execute(message, privateKeys);
```
marekkirejczyk: #mutipleKeys, this and couple more changes in this file

universal-login-sdk/test/events.js
```diff
@@ -34,7 +34,7 @@ describe('SDK - events', async () => {
 
     await sdk.connect(contractAddress);
     const paymentOptions = {...MESSAGE_DEFAULTS, gasToken: mockToken.address};
-    await sdk.addKey(contractAddress, wallet.address, privateKey, paymentOptions);
+    await sdk.addKey(contractAddress, wallet.address, [privateKey], paymentOptions);
```
marekkirejczyk: #mutipleKeys

universal-login-sdk/test/observers/BlockchainObserver.js
```diff
@@ -46,7 +46,7 @@ describe('SDK: BlockchainObserver', async () => {
     const paymentOptions = {...MESSAGE_DEFAULTS, gasToken: mockToken.address};
     const filter = {contractAddress, key: wallet.address};
     await blockchainObserver.subscribe('KeyAdded', filter, callback);
-    await sdk.addKey(contractAddress, wallet.address, privateKey, paymentOptions);
+    await sdk.addKey(contractAddress, wallet.address, [privateKey], paymentOptions);
```
marekkirejczyk: #mutipleKeys, this and couple more changes in this file

universal-login-sdk/test/sdk.js
```diff
@@ -64,69 +64,69 @@ describe('SDK - integration', async () => {
     describe('Execute signed message', async () => {
       it('Should execute signed message', async () => {
         const expectedBalance = (await otherWallet.getBalance()).add(utils.parseEther('0.5'));
-        await sdk.execute({...message, to: otherWallet.address}, privateKey);
+        await sdk.execute({...message, to: otherWallet.address}, [privateKey]);
```
marekkirejczyk: #mutipleKeys, this and couple more changes in this file

universal-login-sdk/types.d.ts
```diff
@@ -12,19 +12,19 @@ declare class EthereumIdentitySDK extends WalletExistenceVerifier  {
 
   create(ensName: string): Promise<[string, string]>;
 
-  addKey(to: string, publicKey: string, privateKey: string, transactionDetails: any, keyPurpose: 1 | 2 | 3): Promise<number>;
+  addKey(to: string, publicKey: string, privateKeys: string[], transactionDetails: any, keyPurpose: 1 | 2 | 3): Promise<number>;
```
marekkirejczyk: #mutipleKeys, this and couple more changes in this file

universal-login-wallet/src/services/TransferService.ts
```diff
@@ -24,7 +24,7 @@ class TransferService {
         data,
         gasToken: tokenAddress
       };
-      await this.sdk.execute(message, this.walletService.userWallet.privateKey);
+      await this.sdk.execute(message, [this.walletService.userWallet.privateKey]);
```
marekkirejczyk: #mutipleKeys

marekkirejczyk:
❤️ Thanks for all the great work! Feedback added in the review.

Keep in mind to keep PRs small. It is better to have 2 small PRs than one medium size 😄 You can build PR at the top of other PRs.

PR #321, opened 2019-03-05
Currencies taken from config

JustynaBroniszewska:


universal-login-wallet/src/services/TokenService.ts
```diff
@@ -0,0 +1,35 @@
...
+      this.tokensDetails[count] = {name, symbol, address: token.address};
+    }
+  }
+
+  getTokenAddress(symbol: string) {
+    for (let count = 0; count < this.tokensDetails.length; count++) {
+      if (this.tokensDetails[count].symbol === symbol) {
+        return this.tokensDetails[count].address;
+      }
+    }
```
marekkirejczyk: Every time you write a for you should consider can it be written using operations on collections, eg:
```
return this.tokensDetails
  .find((token) => token.symbol === symbol)
  .address
```

marekkirejczyk: That probably does not apply to for with `await`'s like the one above :(

universal-login-wallet/src/services/TransferService.ts
```diff
@@ -10,17 +11,18 @@ export interface TransferDetails {
...
 
 class TransferService {
-  constructor(private sdk: UniversalLoginSDK, private walletService: WalletService) {}
+  constructor(private sdk: UniversalLoginSDK, private walletService: WalletService, private tokenService: TokenService) {}
 
   async transferTokens({to, amount, currency} : TransferDetails) {
+    const tokenAddress = this.tokenService.getTokenAddress(currency);
     if (this.walletService.userWallet) {
-      const data = new utils.Interface(MockContract.abi).functions.transfer.encode([to, utils.parseEther(amount)]);
+      const data = new utils.Interface(IERC20.abi).functions.transfer.encode([to, utils.parseEther(amount)]);
```
marekkirejczyk: 👍

universal-login-wallet/test/TokenService.ts
```diff
@@ -0,0 +1,32 @@
...
+    const [wallet] = await getWallets(provider);
+    mockToken = await deployContract(wallet, MockToken, []);
+    tokenService = new TokenService([mockToken.address], provider);
+    symbol = await mockToken.symbol();
+    name = await mockToken.name();
+    await tokenService.start();
+  });
+
+  it('Should fill up token details', async () => {
+    expect(tokenService.tokensDetails[0].name).to.eq(name);
```
marekkirejczyk: I would use here explicit name (minor)

PR #343, opened 2019-03-17
KeyHolder now keeps track on how many keys are in the contract

Giulio2002:
# Summary
In order for the identity to require a certain signatures number in the identity/wallet we need to avoid the user to put a number of required that exceed the actual number of keys present in the keyHolders, to do so we need a variable that increase and decrease along with addition or deletion of keys. i would like to split the addition of a required amount of signatures in the identity in small Pull requests, that's why this Pull Request is very small.

## Checklist
- [X] Change is small and easy to review (please split big changes into multiple PRs)
- [X] Change is consistent with architecture
- [X] Change is consistent with test architecture
- [X] Change is consistent with naming conventions
- [X] New code is covered with tests

### DIFF
--- a/universal-login-contracts/contracts/KeyHolder.sol
+++ b/universal-login-contracts/contracts/KeyHolder.sol
@@ -5,10 +5,12 @@ import "./IKeyHolder.sol";
 contract KeyHolder is IKeyHolder {
     mapping (address => Key) public keys;
 
+    uint _keyCount;
+
     constructor(address _key) public {
         keys[_key].key = _key;
         keys[_key].purpose = MANAGEMENT_KEY;
-
+        _keyCount = 1;
         emit KeyAdded(keys[_key].key,  keys[_key].purpose);
     }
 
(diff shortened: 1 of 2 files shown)


universal-login-contracts/contracts/KeyHolder.sol
```diff
@@ -5,10 +5,12 @@ import "./IKeyHolder.sol";
 contract KeyHolder is IKeyHolder {
     mapping (address => Key) public keys;
```
marekkirejczyk: Make it public, so you don’ need getter later.

universal-login-contracts/contracts/KeyHolder.sol
```diff
@@ -29,6 +31,10 @@ contract KeyHolder is IKeyHolder {
         _;
     }
 
+    function keyCount() public view returns(uint) {
```
marekkirejczyk: Not needed, just make variable public.

universal-login-contracts/contracts/KeyHolder.sol
Giulio2002: Curiosity: why in ERC1077 _lastNonce is a private variable with a getter and not simply a public variable?

marekkirejczyk: Probably historical reasons. `public uint lastNonce;` would be good.

PR #351, opened 2019-03-19
Support for nested ENS domain

Amxx:
Current splitting method assumes `label.domain.eth` structure to extract label and domain. This only work because ethers.js autocompletes domain to domain.eth. It will not work with other tld's and with subdomain.

To support `user.subdomain.domain.xyz`, the splitting should be done at the first `.` only, so that `label = user` and `node=subdomain.domain.xyz`. This is what this PR implements

### DIFF
--- a/universal-login-relayer/lib/services/ensService.js
+++ b/universal-login-relayer/lib/services/ensService.js
@@ -23,16 +23,8 @@ class ENSService {
     return this.domainsInfo[domain] || null;
   }
 
-  get2ndLevelDomainForm(ensName) {
-    const labels = ensName.split('.');
-    const {length} = labels;
-    const label = labels.slice(0, length - 2).join('.');
-    const domain = labels.slice(length - 2, length).join('.');
-    return [label, domain];
-  }
-
   argsFor(ensName) {
-    const [label, domain] = this.get2ndLevelDomainForm(ensName);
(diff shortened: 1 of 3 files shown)


Amxx:
Failure of the get2ndLevelDomainForm test is expected since I replaced this function. Can someone explain the reason that led to this function being here in the first place ?

marekkirejczyk:
Hi @Amxx,

I think you just implemented similar functionality in a much shorter way :)
I would propose to move implementation to `domain.split(/\.(.*)/)` to the function (so that test passes)
I would also consider renaming function as I think it does something else than the name suggest. The name could be `parseDomain` or `splitDomin` or sth like this.

I guess complicated implementation is here for historical reasons, but not sure anymore.

Marek

Amxx:
If you look at the tests, `john.and.marek.universal-id.eth` is supposed to be slip into `john.and.marek` and `universal-id.eth`. Also `universal-id.eth` is expected to turn into `` and `universal-id.eth`

My split function will return `john`, `and.marek.universal-id.eth` in the first case and `universal-id`, `eth` in the second case

marekkirejczyk:
Fair enough, shall we fix tests?

marekkirejczyk:
Fix and merged in follow-up PR #362

PR #411, opened 2019-04-05
[WIP] Debugging the simplified version of contract KeyHolder

tkazana-eth:
Changes:

* timeout to 50 min
* simplified spec

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,12 +124,14 @@ jobs:
           sudo make link
       - setup_remote_docker:
           docker_layer_caching: true
-      - run: |
-          export PATH=$PATH:/usr/local/bin
-          export KLAB_EVMS_PATH=/home/circleci/project/klab/evm-semantics
-          export TMPDIR=/tmp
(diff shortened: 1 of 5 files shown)


tkazana-eth:
Some proofs _sometimes_ are not accepted. Probably the reason is connected to resources of CI (all are accepted on my MacBook Air). Will examine the problem after my vacation.

tkazana-eth:
I have dug in the issue of the last comment. So: error 137 (that we _sometimes_ receive) is connected to 'out-of-memory' issue on CircleCI.
It means that the last simplification (proposed by Marek) is a trade-off: the spec is simpler, but the proof is more exhausting because the IFF clause is longer and needs more analysis from the prover...

For now: the last stable version of FV is the one BEFORE the last commit (with SafeMath etc). Maybe we should revert and later discuss the strategy of the whole FV process.

marekkirejczyk:
Build fails. Closing for now. Waiting for @tkazana-eth to come back from Holidays to fix formal verification CI.

PR #447, opened 2019-04-29
Updated implementation and spec for KeyHolder

tkazana-eth:
# Summary
* SafeMath (from openzeppelin) added
* simple requirements added (i.e. key != msg.sender) => it allowed to simplify the ACT spec of functions: addKey and removeKey

## technical modifications
* the whole process is now done with Yarn and Waffle 2.0.11. It makes the whole building process of Universal Login more unified.

## Checklist


tkazana-eth:
The update from Waffle 2.0.5 to 2.0.11 is not carried out for all package.json filea in all subdirectories. It is NOT updated for sdk and relayer because the update makes the tests fail. It should be examined in the future.

PR #451, opened 2019-04-30
added login interface to compound

Giulio2002:
# Summary
added login interface to compound

## Checklist


marekkirejczyk:
Thank you for the #PR. We discussed it internally and we are not sure if that is a good direction. I think having a separate screen is not optimal user experience. I would rather see compound as "saving account" on the list of accounts.  I will ask our design team to propose sth.

Giulio2002:
@

> Thank you for the #PR. We discussed it internally and we are not sure if that is a good direction. I think having a separate screen is not optimal user experience. I would rather see compound as "saving account" on the list of accounts. I will ask our design team to propose sth.

sorry for my early inactivity on this, i have been really busy lately. so, how do you think that compound should be structured? do you mean creating a compound wallet along with the creation of a normal wallet or making the universalLoginWallet handle more than one Wallet contract and put the compound wallet there?

Many thanks in advance.

JustynaBroniszewska:
We are still working with our design team. We will let you know as soon as we have update.

PR #478, opened 2019-05-11
Small gas improvement in ERC1077

canepat:
# Summary
Small gas improvement in ERC1077.

## Checklist
- [x] Change is small and easy to review

### DIFF
--- a/universal-login-contracts/contracts/ERC1077.sol
+++ b/universal-login-contracts/contracts/ERC1077.sol
@@ -118,7 +118,7 @@ contract ERC1077 is KeyHolder, IERC1077, IERC721Receiver {
         /* solium-disable-next-line security/no-call-value */
         (success, _data) = to.call.value(value)(data);
         bytes32 messageHash = calculateMessageHash(address(this), to, value, data, nonce, gasPrice, gasToken, gasLimit, operationType);
-        emit ExecutedSigned(messageHash, lastNonce-1, success);
+        emit ExecutedSigned(messageHash, nonce, success);
         uint256 gasUsed = startingGas.sub(gasleft());
         refund(gasUsed, gasPrice, gasToken);
         return messageHash;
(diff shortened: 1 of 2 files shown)


universal-login-contracts/contracts/ERC1077.sol
```diff
@@ -118,7 +118,7 @@ contract ERC1077 is KeyHolder, IERC1077, IERC721Receiver {
         /* solium-disable-next-line security/no-call-value */
         (success, _data) = to.call.value(value)(data);
         bytes32 messageHash = calculateMessageHash(address(this), to, value, data, nonce, gasPrice, gasToken, gasLimit, operationType);
-        emit ExecutedSigned(messageHash, lastNonce-1, success);
```
marekkirejczyk: Why would we like to increase nonce? Note that we increase the nonce in line 114, therefore `nonce-1` equals to actual nonce.

canepat: I just substituted `lastNonce-1` with `nonce`, which happen to be the same value as per lines 112 and 114, for performance reasons: saving 1 SLOAD and 1 SUB opcodes

marekkirejczyk: That would, unfortunately, change the semantics. 
We moved `lastNonce` incrementation before the call to avoid re-entry.

NoahMarconi: @marekkirejczyk  this change doesn't remove the `lastNonce` increment, it's still there on the same line it always was.

All this change does is make use of the `nonce` argument already passed into the function instead of inlining an unnecessary `lastNonce-1` as `nonce` is already in scope and is equal to `lastNonce-1`

marekkirejczyk: Good point. My bad.

PR #532, opened 2019-05-27
Reduce number of arguments when creating a proxy (registrar and resolver)

JustynaBroniszewska:


marekkirejczyk:
Ok, so arguments are actually cheaper than external calls. Lesson learned. Closing for now.

PR #548, opened 2019-06-11
Blank factory contract

JustynaBroniszewska:


universal-login-contracts/contracts/Factory.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity ^0.5.2;
+
+contract Factory {
```
marekkirejczyk: `ProxyCounterfactualFactory`?

PR #576, opened 2019-06-27
Basic 'connect new device' doc

JustynaBroniszewska:


JustynaBroniszewska:
* old device generate temporary key pair
* private key gets transferred to new device
* new device encrypts new public key using a temporary private key
* sends via relayer
* on successful decription, the old device sends meta-transaction to relayer to add the new public key to wallet smart contract

 .. image:: static/connect/solution-2.png

PR #577, opened 2019-06-27
Improve copy

xitronix:
Changes required by nondeterministic build failings

### DIFF
--- a/universal-login-commons/lib/utils/copy.ts
+++ b/universal-login-commons/lib/utils/copy.ts
@@ -1,5 +1,9 @@
 export function copy(id : string) {
-  const content = document.getElementById(id) as HTMLInputElement;
-  content.select();
-  document.execCommand('copy');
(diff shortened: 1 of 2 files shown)


universal-login-commons/lib/utils/copy.ts
```diff
@@ -1,5 +1,9 @@
 export function copy(id : string) {
-  const content = document.getElementById(id) as HTMLInputElement;
-  content.select();
-  document.execCommand('copy');
+  const content = document.getElementById(id);
+  if (content instanceof HTMLInputElement) {
+    content.select();
+    document.execCommand('copy');
+  } else {
+    throw Error(`Element (#${id}) does not exist`);
```
marekkirejczyk:
```ts
ensure(content instanceof HTMLInputElement, Error, 'Element (#${id}) does not exist')
```

PR #594, opened 2019-07-02
Segregate services/* 

xitronix:


universal-login-wallet/src/ui/react/Login/ApproveScreen.tsx
```diff
@@ -7,7 +7,7 @@ import { Link } from 'react-router-dom';
 import {useServices} from '../../hooks';
 
 const ApproveScreen = () => {
-  const {walletFormatter} = useServices();
+  const {walletPresenter: walletFormatter} = useServices();
```
marekkirejczyk: Why not `const {walletPresenter} = useServices()`

PR #611, opened 2019-07-04
Move authRequest creation to httpUtils

gasiortomasz:


universal-login-relayer/lib/utils/httpUtils.ts
```diff
@@ -0,0 +1,17 @@
+import {Request} from 'express';
+import geoip from 'geoip-lite';
+import moment from 'moment';
+
+
+export const getDeviceInfo = (req: Request) => {
```
marekkirejczyk: Add a return type to function: `DeviceInfo` (from `commons`).

PR #664, opened 2019-07-23
Improve authorisations forwarding in AuthorisationsObserver.

gasiortomasz:


universal-login-sdk/lib/core/observers/AuthorisationsObserver.ts
```diff
@@ -22,8 +22,9 @@ class AuthorisationsObserver extends ObserverRunner {
 
     if (!deepEqual(authorisations, this.lastAuthorisations)) {
       this.lastAuthorisations = authorisations;
+      const authorisationsCopy = authorisations.map((a: Notification) => ({...a}));
```
xitronix: Deconstructing will just shallow copy obj. I guess deep copy will be safer? :
const authorisationsCopy = JSON.parse(JSON.stringify(authorisations));

PR #674, opened 2019-07-24
Wallet selector structure change

alexcherman:


universal-login-react/src/ui/styles/walletSelector.css
```diff
@@ -40,11 +39,13 @@
   border: none;
 }
 
-.universal-login .suggestions-item-btn-text {
+.suggestions-item-btn-text {
```
marekkirejczyk: `.universal-login` needs to stay here

PR #677, opened 2019-07-24
Documentation: Update relayer section

JustynaBroniszewska:


marekkirejczyk:
We need getStatus documentation and we can finish with docs for now

PR #709, opened 2019-07-31
Exclude build and .cache folders in VSC search tool

gasiortomasz:


JustynaBroniszewska:
I am not sure if this should be a workspace setting   🤔  I would add it as one of the user settings

PR #746, opened 2019-08-07
Introduce emoji fonts

xitronix:


universal-login-react/src/core/utils/emojiCodePoint.ts
```diff
@@ -0,0 +1,962 @@
+export const getEmojiCodePoint = (code: number): number => {
```
marekkirejczyk: `getCodePointFromSecurityCode`

PR #748, opened 2019-08-07
Change recieve modal design

alexcherman:


universal-login-wallet/src/ui/react/Modals/ModalRequest.tsx
```diff
@@ -1,41 +1,52 @@
...
+    return (
+      <div className={`hint ${color}`}>
+        <p className="hint-text">{children}</p>
+        <button onClick={() => setHintVisibility(false)} className="hint-btn">Dismiss</button>
+      </div>
+    );
+  }
+
+  return null;
+};
```
marekkirejczyk: 👍 small nice component, still I would extract it to separate file ;P

PR #750, opened 2019-08-07
Add linter rule + Fill missing awaits

gasiortomasz:


universal-login-relayer/lib/core/services/MessageHandler.ts
```diff
@@ -61,8 +61,8 @@ class MessageHandler {
     return this.pendingMessages.getStatus(messageHash);
   }
 
-  stop() {
-    this.queueService.stop();
+  async stop() {
+    await this.queueService.stop();
   }
```
marekkirejczyk: Simpler:
```ts
stop() {
  return  this.queueService.stop();
}
```

PR #758, opened 2019-08-09
Show message on creation if ens name taken

xitronix:


universal-login-react/src/ui/WalletSelector/WalletSelector.tsx
```diff
@@ -55,6 +61,7 @@ export const WalletSelector = ({onCreateClick, onConnectClick, sdk, domains, act
             placeholder="bob.example.eth"
             autoFocus
           />
+          {isNameTaken() && <div>Name is already taken</div>}
```
JustynaBroniszewska: extract `renderNameTaken()`

PR #768, opened 2019-08-13
Introduce U component

gasiortomasz:
Added U component (as LogoButton).

In order to make this component work, there is additional component/service that creates client instance (as CreateRandomInstance - it deploys contract and stores ApplicationWallet that is passed to U component in props)


universal-login-react/src/ui/hooks/useAsyncEffect.ts
```diff
@@ -0,0 +1,12 @@
+import {useEffect, DependencyList} from 'react';
+
+type Unsubscribe = () => void;
+
+export const useAsyncEffect = (asyncCallback: () => Promise<Unsubscribe>, deps?: DependencyList) => {
```
marekkirejczyk: Duplicate?
https://github.com/UniversalLogin/UniversalLoginSDK/blob/d62bdc8a0a2e3880791dd530f97dae1424de21e2/universal-login-wallet/src/ui/hooks/useAsyncEffect.ts

PR #769, opened 2019-08-13
Connect with Ethereum Button

alexvandesande:
First of all sorry for the double quotes and other formatting, I'm figuring out some stuff with Lint. Anyways here are the changes in this button:

* Starts with a "connect with ethereum" button that checks for a web3 browser
* If it doesn't have one, or if the user denies it, then proceeds with normal universal login flow
* changed colors from UL purple to Classic Grey & White

This branch should be considered a demo only, there are some UI issues with Safari and probably other problems.


marekkirejczyk:
Closing. Will work on that in a separate repository.

PR #791, opened 2019-08-15
Improve token value aggregation

gasiortomasz:


universal-login-commons/lib/core/services/TokensValueConverter.ts
```diff
@@ -36,7 +36,9 @@ export class TokensValueConverter {
 
   addBalances(totalBalances: CurrencyToValue, toAddBalances: CurrencyToValue) {
     for (const key in totalBalances) {
-      totalBalances[key as ObservedCurrency] += toAddBalances[key as ObservedCurrency];
+      if (key in toAddBalances && typeof toAddBalances[key as ObservedCurrency] === 'number') {
+        totalBalances[key as ObservedCurrency] += toAddBalances[key as ObservedCurrency];
+      }
```
marekkirejczyk: Would be good to add tests (next pr/next week)?

marekkirejczyk: Also probably sth wrong is with types here.

PR #795, opened 2019-08-15
Make same icons for connected devices

xitronix:


universal-login-wallet/src/core/presenters/WalletPresenter.ts
```diff
@@ -15,7 +15,7 @@ export class WalletPresenter {
 
   getContractAddress(): string {
     ensure(!!this.walletService.applicationWallet, ApplicationWalletNotFound);
-    return this.walletService.applicationWallet!.contractAddress;
+    return this.walletService.applicationWallet!.contractAddress.toLowerCase();
```
gasiortomasz: +1

gasiortomasz: It should be in WalletService while creating ApplicationWallet, no?

PR #832, opened 2019-08-19
Improve emoji

xitronix:


universal-login-react/src/ui/Notifications/EmojiPlaceholders.tsx
```diff
@@ -14,7 +14,7 @@ export const EmojiPlaceholders = ({code, onEmojiClicked, maxLength}: EmojiPlaceh
     const emojis = [];
 
     for (let i = 0; i < maxLength; i++) {
-      if (code[i]) {
+      if (typeof code[i] === 'number') {
```
marekkirejczyk: Ugly. Can we give it a type? Why would it not be a number?

xitronix: 1. We use **i** for indexing React components. 
2. `codes.length` can have `length !==0 && length !== maxLength`. This component use  placeholders for button which appears when you click some code in FakeCodes. 
3. code[i] can be falsy value (`code[i]=0 || undefined`) that's why wee need check type or `code[i] !== undefined` to know what should be render `<li><button /></li>` or just `<li />`

universal-login-react/src/ui/Notifications/EmojiPlaceholders.tsx
```diff
@@ -9,22 +9,22 @@ interface EmojiPlaceholdersProps {
...
+    <li key={i}>
+      <button onClick={() => onEmojiClicked(i)}>
+        <Emoji code={code} />
+      </button>
+    </li>
+  ));
+
 export const EmojiPlaceholders = ({code, onEmojiClicked, maxLength}: EmojiPlaceholdersProps) => {
+
   const renderSelectedEmojis = () => {
```
JustynaBroniszewska: My suggestion: 
```ts
const renderSelectedEmojis = () => {
    const items = code.map((code: number, i: number) => getEmojiButton(code, i, onEmojiClicked));
    for (let i = code.length; i < maxLength; i++) {
      items.push(getEmojiPlaceholder(i));
    }
    return items;
  };
```

PR #839, opened 2019-08-21
Move http and fetch from SDK to commons

JustynaBroniszewska:


universal-login-commons/package.json
```diff
@@ -29,6 +30,7 @@
   "dependencies": {
     "ethers": "^4.0",
     "lodash.clonedeep": "^4.5.0",
+    "node-fetch": "^2.6.0",
```
marekkirejczyk: Not sure if it is a good idea to add node-fetch to commons dependencies. It is a node-specific library. Will most likely cause troubles on front-end.

PR #844, opened 2019-08-21
Add services to store users wallet in local storage

Marik-D:
Contains #855


universal-login-wallet/test/integration/helpers/chaiConfig.ts
```diff
@@ -0,0 +1,6 @@
+import chai from 'chai';
+import sinonChai from 'sinon-chai';
+import {solidity} from 'ethereum-waffle';
+
+chai.use(solidity);
+chai.use(sinonChai);
```
marekkirejczyk: That should probably be a separate PR.

marekkirejczyk: Is it used anywhere?

universal-login-wallet/src/ui/react/App.tsx
```diff
@@ -19,7 +19,18 @@ import {ConnectionNotificationScreen} from './ConnectAccount/ConnectionNotificat
...
+  const {walletService, walletStorageService, sdk} = useServices();
+  const [appReady, setAppReady] = useState(false);
+
+  useLayoutEffect(() => {
+    walletStorageService.load();
+    setAppReady(true);
+  }, []);
+
+  if (!appReady) {
+    return null;
```
marekkirejczyk: Shall we do `return 'Loading...';` instead?

universal-login-wallet/src/core/services/StorageService.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export class StorageEntry<T> {
+  constructor(
+    private key: string,
+    private sanitizer: Sanitizer<T>,
+    private storageService: StorageService,
+  ) {
+  }
+
```
marekkirejczyk: What's with the newlines? We should always use one empty line between methods.

universal-login-wallet/src/core/services/StorageService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  get(): T | null {
+    const json = this.storageService.get(this.key);
+    if (!json) {
+      return null;
+    }
+
```
marekkirejczyk: Empty newline in a method? Meh.

universal-login-wallet/src/core/services/StorageService.ts
```diff
@@ -0,0 +1,35 @@
...
+  get(key: string): string | null {
+    return localStorage.getItem(key);
+  }
+
+  set(key: string, value: string) {
+    localStorage.setItem(key, value);
+  }
+}
+
+export class StorageEntry<T> {
```
marekkirejczyk: Abstraction level 10000 👍 😄

universal-login-wallet/src/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,40 @@
...
+import {WalletService} from '@universal-login/sdk';
+import {ApplicationWallet, ensure} from '@universal-login/commons';
+import {asObject, asString, Sanitizer} from '@restless/sanitizers';
+
+export class WalletStorageService {
+  private storage: StorageEntry<ApplicationWallet>;
+
+  constructor(
+    private walletService: WalletService,
+    storageService: StorageService,
```
marekkirejczyk: We should make a maximum length of a line of 120 characters. Simple function headers take to many lines.

universal-login-wallet/src/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,40 @@
+import {StorageEntry, StorageService} from './StorageService';
+import {WalletService} from '@universal-login/sdk';
+import {ApplicationWallet, ensure} from '@universal-login/commons';
+import {asObject, asString, Sanitizer} from '@restless/sanitizers';
+
+export class WalletStorageService {
+  private storage: StorageEntry<ApplicationWallet>;
+
+  constructor(
+    private walletService: WalletService,
```
marekkirejczyk: Shouldn't dependency be another way around? i.e. WalletService depend on StorageService?

universal-login-wallet/src/ui/react/ConnectAccount/ConnectWithPassphrase.tsx
```diff
@@ -11,13 +11,14 @@ export const ConnectWithPassphrase = ({name}: ConnectWithPasssphraseProps) => {
...
   const [errorMessage, setErrorMessage] = useState('');
-  const {walletService} = useServices();
+  const {walletService, walletStorageService} = useServices();
   const {history} = useRouter();
 
   const onRecoveryClick = async () => {
     setIsLoading(true);
     try {
       await walletService.recover(name, code);
+      walletStorageService.save();
```
marekkirejczyk: Should be just one line: 
await walletService.recover(name, code);
#reversedependencies

universal-login-wallet/src/ui/react/CreateAccount/CreateAccount.tsx
```diff
@@ -23,6 +23,7 @@ export const CreateAccount = ({location}: CreateAccountProps) => {
     modalService.showModal('waitingForDeploy');
     await deploy(name, defaultDeployOptions.gasPrice.toString());
     walletService.setDeployed(name);
+    walletStorageService.save();
```
marekkirejczyk: Should be just one line. #reversedependencies

universal-login-wallet/src/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,40 @@
...
+    const wallet = this.walletService.applicationWallet as ApplicationWallet;
+    this.storage.set(wallet);
+  }
+
+  load() {
+    const wallet = this.storage.get();
+    if (wallet === null) {
+      return false;
+    }
+    this.walletService.connect(wallet);
```
marekkirejczyk: #reversedependencies

universal-login-wallet/src/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,40 @@
...
+      'wallet',
+      asApplicationWallet,
+      storageService,
+    );
+  }
+
+  save() {
+    ensure(this.walletService.isAuthorized(), Error, 'User must be authorized');
+    const wallet = this.walletService.applicationWallet as ApplicationWallet;
+    this.storage.set(wallet);
```
marekkirejczyk: #reversedependencies

universal-login-wallet/test/integration/helpers/chaiConfig.ts
Marik-D: Yep, it's used in `LoginServices` test

universal-login-wallet/src/ui/react/App.tsx
Marik-D: This will result in a flash of text in this case. I've only added it here since the wallet service is queried directly, which means that it wouldn't update the component by itself. So we need a state just to trigger a re-render.

universal-login-wallet/src/core/services/StorageService.ts
Marik-D: `StorageEntry` is just a nice interface that bundles up together a storage key and a sanitizer for a specific value to prevent mixups and type-unsafe value casts.

universal-login-wallet/src/core/services/WalletStorageService.ts
Marik-D: In my opinion, even with 2 parameters, this variant is easier to read, since every parameter occupies a single line, it's easy to distinguish names from types. And you can just skim through it to know what that constructor takes.

universal-login-wallet/src/ui/react/App.tsx
Marik-D: We should think about a way to add an ability to watch for changes in the wallet service.

universal-login-wallet/src/core/services/WalletStorageService.ts
Marik-D: Yeah, I can add a wallet storage as a parameter to wallet service that defaults to in-memory storage

universal-login-wallet/src/ui/react/App.tsx
marekkirejczyk: It is on the roadmap.

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -8,11 +8,21 @@ type WalletState = 'None' | 'Future' | 'Deployed';
 
 type WalletFromBackupCodes = (username: string, password: string) => Promise<Wallet>;
 
+export interface WalletStorage {
+  load(): ApplicationWallet | null;
+
+  save(wallet: ApplicationWallet | null): void;
+}
```
marekkirejczyk: extract file?

PR #849, opened 2019-08-22
Contracts: Export test utils and rewrite test to ts

JustynaBroniszewska:


universal-login-contracts/testutils/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "main": "../dist/test/fixtures/index.js",
+  "types": "../dist/test/fixtures/index.d.ts"
+}
```
marekkirejczyk: Not sure what does this file do?

universal-login-contracts/test/fixtures/index.ts
```diff
@@ -0,0 +1 @@
+export {setupMasterWithRefundAndFactory} from './walletContract';
```
marekkirejczyk: Why do we need this file?

PR #852, opened 2019-08-22
Improve dropdown

xitronix:


universal-login-sdk/lib/integration/ethereum/TokensDetailsStore.ts
```diff
@@ -6,10 +6,7 @@ export class TokensDetailsStore {
   constructor(private tokenDetailsService: TokenDetailsService, private tokensAddresses: string[]) {}
 
   async fetchTokensDetails() {
-    for (const address of this.tokensAddresses) {
-      const details = await this.tokenDetailsService.getTokenDetails(address);
-      this.tokensDetails.push(details);
-    }
+    this.tokensDetails = await Promise.all(this.tokensAddresses.map((address) => this.tokenDetailsService.getTokenDetails(address)));
```
marekkirejczyk: `Promise.all` -> almost always bad idea. If someone feels like it, we could create util, i.e.:
```ts
Promise.batched(5, Promise[])
` to improve performance, but it still would need to be low concurrency rate and configurable.

File: universal-login-react/src/ui/Transfer/Amount/TransferDropdown.tsx [hunk 2]
```diff
@@ -29,43 +28,33 @@ export const TransferDropdown = ({sdk, ensName, currency, setCurrency}: Transfer
 
   const iconForToken = (symbol: string) => symbol === 'ETH' ? ethIcon : daiIcon;
 
+  const getBalanceOf = (symbol: string) => {
```
marekkirejczyk: Extract a utility method and deduplicate.

marekkirejczyk:
Needs rebase @xitronix

PR #863, opened 2019-08-23
Improve ObserverRunner

xitronix:
That's PR reduce wallet's test time too

### DIFF
--- a/universal-login-sdk/lib/core/observers/ObserverRunner.ts
+++ b/universal-login-sdk/lib/core/observers/ObserverRunner.ts
@@ -6,6 +6,7 @@ abstract class ObserverRunner {
   protected state: ObserverRunnerState = 'stop';
   abstract async tick(): Promise<void>;
   step = 1000;
+  timeout: any = null;
 
   async loop() {
     if (this.state === 'stop') {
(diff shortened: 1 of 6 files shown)


universal-login-sdk/lib/core/observers/ObserverRunner.ts
```diff
@@ -28,10 +29,11 @@ abstract class ObserverRunner {
 
   stop() {
     this.state = 'stop';
+    clearTimeout(this.timeout);
   }
 
   async finalizeAndStop() {
-    this.state = 'stopping';
+    this.state = this.isRunning() ? 'stopping' : 'stop';
```
marekkirejczyk: This is probably a bug, i.e.:
`if (state == 'stopping')` `finalizeAndStop()` will change it to `stop` while it should not.

xitronix: Implementation of isRunnning bellow: 
``` ts 
protected isRunning() {
    return this.state !== 'stop';
 }
```
So if state === 'stopping' finalizeAndStop `isRunning()` will be true, so finalizeAndStop will set state to 'stopping'

marekkirejczyk: Ok, so it is a bad getter. IsRunning should returninfo if observer is in state running not running or stopping

xitronix: So I will update it and change condition

PR #866, opened 2019-08-26
Introduce securityCodes for keyboard mode

gasiortomasz:

PR #869, opened 2019-08-26
Add missing default class in react emoji components

gasiortomasz:


JustynaBroniszewska:
#869 and #866 don't depend on each other so I don't understand why you pushed it together.

PR #903, opened 2019-08-30
React: Refactor Assets and Asset and use it in wallet

gasiortomasz:


universal-login-wallet/src/ui/react/Home/Asset.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {TokensPrices} from '@universal-login/commons';
+import {useServices} from '../../hooks';
+
+export interface AssetProps {
+  name: string;
+  symbol: string;
+  balance: string;
+}
+
+export const Asset = ({name, symbol, balance}: AssetProps) => {
```
JustynaBroniszewska: Similar to [Asset.tsx](https://github.com/UniversalLogin/UniversalLoginSDK/blob/master/universal-login-react/src/ui/commons/Asset.tsx). Let's import it from `@universal-login/react`.

JustynaBroniszewska: I found one more similar component [TransferDropdownItem](https://github.com/UniversalLogin/UniversalLoginSDK/blob/master/universal-login-react/src/ui/Transfer/Amount/TransferDropdownItem.tsx). 

We should write more generic Asset component and export it from `@universal-login/react`.

gasiortomasz: True, wasn't aware of them

PR #913, opened 2019-09-02
Improve disconnect flow

gasiortomasz:


universal-login-wallet/src/core/services/WalletStorageService.ts
```diff
@@ -21,6 +21,10 @@ export class WalletStorageService implements WalletStorage {
   save(wallet: ApplicationWallet | null): void {
```
marekkirejczyk: Unnecessary `| null`

PR #940, opened 2019-09-05
Set up web3 package

Marik-D:


marekkirejczyk:
Would be good to add configuration to .circleci

PR #942, opened 2019-09-05
Introduce script for register .eth domain

JustynaBroniszewska:


universal-login-ops/src/ENS/DomainRegistrar.ts
```diff
@@ -59,6 +74,17 @@ class DomainRegistrar extends ENSRegistrarBase {
     saveVariables(filename, this.variables);
     this.log(`${filename} file updated.`);
   }
+
+  async registerEthDomain(label : string) {
+    const tld = 'eth';
```
marekkirejczyk: move to arg

PR #954, opened 2019-09-06
Adopt TokenDetailsService to fetch DAI name and symbol

JustynaBroniszewska:


universal-login-commons/lib/integration/ethereum/TokenDetailsService.ts
```diff
@@ -1,12 +1,17 @@
...
 
-const tokenAbi = [
+const tokenAbiString = [
   'function name() public view returns (string)',
   'function symbol() public view returns (string)'
 ];
 
+const tokenAbiBytes32 = [
+  'function name() public view returns (bytes32)',
+  'function symbol() public view returns (bytes32)'
```
marekkirejczyk: But why? Is DAI incompatible with ERC20?

PR #969, opened 2019-09-09
Remove nonce from executeSigned function

JustynaBroniszewska:


universal-login-contracts/test/contracts/Executor/Executor.ts
```diff
@@ -128,13 +128,14 @@ describe('CONTRACT: Executor - main', async  () => {
...
-          .to.be.revertedWith('Invalid nonce');
+          .to.be.revertedWith('Invalid signature');
       });
 
       it('nonce too high', async () => {
         msg = {...transferMessage, from: walletContract.address, nonce: 2};
+        signature = calculateMessageSignature(managementKeyPair.privateKey, msg);
         await expect(walletContract.executeSigned(...getExecutionArgs(msg), signature, DEFAULT_PAYMENT_OPTIONS_NO_GAS_TOKEN))
-          .to.be.revertedWith('Invalid nonce');
+          .to.be.revertedWith('Invalid signature');
```
marekkirejczyk: `Invalid signature or nonce`

universal-login-contracts/contracts/wallet/Executor.sol
```diff
@@ -19,11 +19,11 @@ contract Executor {
     }
 
     function etherRefundCharge() public pure returns(uint) {
-        return 14000;
+        return 14500;
```
marekkirejczyk: Is it actually more expensive without `nonce` than with `nonce` as an argument?

JustynaBroniszewska: Yes, we execute 2 more operations: `lastNonce.sub(1)` in lines 116 and 117

PR #1033, opened 2019-09-17
Wallet: Fix error after clicking previous page

cehali:
# Summary
When creating an account, clicking the previous page and accepting terms once again generated an error with config variables. Simple fix, the current behavior of the application can be discussed.

Fixes: 1

## Checklist


universal-login-wallet/src/ui/react/Modals/Modal.tsx
```diff
@@ -38,7 +38,7 @@ const Modal = () => {
         </ModalWrapperClosable>
       );
     case 'topUpAccount':
-      return (
+      return config ? (
```
marekkirejczyk: Why would config be null?

cehali: When we click the previous button on topUp modal, we see terms. Accepting them moves us to topUp modal once again (skipping choosing name). Modal component has the function 'useRelayerConfig' which uses 'useAsync'. When we click the previous button and then go to topUp modal, modal component tries to get config but 'useRelayerConfig' is not resolved yet and this is why we have an undefined config for short period of time. 'useAsync' rerenders the page and 'useRelayerConfig' is almost instant so the user is not aware of rerendering.

Another option is to wait until 'config' isn't undefined and render some loader but a short period of time, which would be more like blink than loading.

marekkirejczyk: Let's use cache instead

universal-login-wallet/src/ui/createServices.ts
```diff
@@ -44,6 +44,7 @@ export const createServices = (config: Config, overrides : Overrides = {}) => {
     walletService,
     walletPresenter,
     start: () => sdk.start(),
+    getRelayerConfig: () => sdk.getRelayerConfig()
```
marekkirejczyk: ech... what is that? is that a service?

universal-login-wallet/src/ui/hooks/useRelayerConfig.ts
```diff
@@ -1,8 +0,0 @@
-import {useAsync} from '@universal-login/react';
-import {useServices} from './useServices';
-
-export const useRelayerConfig = () => {
-  const {sdk} = useServices();
-  const [config] = useAsync(async () => sdk.getRelayerConfig(), []);
```
marekkirejczyk: This is actually cool, I would leave it, just update to use `configService`

universal-login-wallet/src/core/services/ConfigService.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class ConfigService {
+  private sdk: UniversalLoginSDK;
+  public relayerConfig?: PublicRelayerConfig;
+
+  constructor(sdk: UniversalLoginSDK) {
+    this.sdk = sdk;
+  }
+
+  async setRelayerConfig() {
```
marekkirejczyk: `fetchRelayerConfig`

universal-login-wallet/src/core/services/ConfigService.ts
```diff
@@ -0,0 +1,21 @@
...
+    if (!this.relayerConfig) {
+      this.relayerConfig = await this.sdk.getRelayerConfig();
+    }
+  }
+
+  getRelayerConfig() {
+    return this.relayerConfig;
```
marekkirejczyk:
```ts
async getRelayerConfig() {
  await fetchRelayerConfig();
  return this.relayerConfig;
}
```

universal-login-wallet/src/ui/react/Modals/Modal.tsx
```diff
@@ -2,18 +2,20 @@ import React, {useContext} from 'react';
...
 import {WalletModalContext} from '../../../core/entities/WalletModalContext';
 import {ConnectionNotificationModal} from '../ConnectAccount/ConnectionNotificationModal';
 
 const Modal = () => {
   const modalService = useContext(WalletModalContext);
-  const {walletPresenter, walletService, sdk} = useServices();
-  const config = useRelayerConfig();
+  const {walletPresenter, walletService, sdk, configService} = useServices();
+  useAsync(async () => configService.setRelayerConfig(), []);
+  const relayerConfig = configService.getRelayerConfig();
```
marekkirejczyk: just use `useRelayerConfig`

PR #1055, opened 2019-09-19
Relayer: Use provider in validators

rzadp:
https://github.com/UniversalLogin/UniversalLoginSDK/pull/1054#discussion_r326109566


rzadp:
@marekkirejczyk The tests are failing on `yarn codechecks`, seems like the service is temporarily unavailable? `master` job fails too

rzadp:
@marekkirejczyk the tests are passing now and this is ready for review

PR #1063, opened 2019-09-20
Add network property to relayerConfig

xitronix:


universal-login-relayer/lib/config/config.prod.ts
```diff
@@ -6,6 +6,7 @@ export const config: Config =  Object.freeze({
   jsonRpcUrl: getEnv('JSON_RPC_URL', ''),
   port: getEnv('PORT', ''),
   privateKey: getEnv('PRIVATE_KEY', ''),
+  network: getEnv('NETWORK', ''),
   chainSpec: Object.freeze({
```
marekkirejczyk: idea: shell we add `name` to `chainspec` instead?

xitronix: 👍  Already exists, so we should just set it from env in prod mode

marekkirejczyk: Cool, but we should add information in all the different files (test, dev).

PR #1101, opened 2019-09-25
Wallet: Add max button functionality to transfers

cehali:
# Summary
Add functionality to Max button in Transfer Modal. Unfortunately, the estimation of gas used is not 100% accurate which leads to the situation that after sending max amount the user would still have a small amount of ether/token. Additionally, even in the test, estimations are not always the same which means test cannot be consistent, this is why I used greater/lesser than in test. 

Fixes: 0

## Checklist


marekkirejczyk:
Closing, please reopen when ready.

PR #1105, opened 2019-09-25
Typo

rzadp:


universal-login-relayer/lib/integration/sql/services/QueueSQLStore.ts
```diff
@@ -14,7 +14,7 @@ export default class QueueSQLStore implements IQueueStore {
     await this.database
       .insert({
         hash: messageHash,
-        type: 'Mesasge',
```
marekkirejczyk: Note, in the future 
- we should add migration for that (to rename existing records in db)
- We should add some type safety to this

PR #1106, opened 2019-09-25
Generic IQueueStore

rzadp:
Generic interface for queue will allow to re-use it for new types of queues.


universal-login-relayer/lib/core/services/messages/IQueueStore.ts
```diff
@@ -1,11 +1,10 @@
 import {SignedMessage} from '@universal-login/commons';
 import {QueueItem} from '../../models/messages/QueueItem';
 
-export interface IQueueStore {
-  add: (message: SignedMessage) => Promise<string>;
-  getNext: () => Promise<QueueItem | undefined>;
-  remove: (messageHash: string) => Promise<void>;
+export interface IQueueStore <A, B, C> {
```
marekkirejczyk: A, B, and C seems to be the same type

PR #1159, opened 2019-09-30
On connection success show devices

xitronix:


universal-login-react/src/App.tsx
```diff
@@ -124,6 +124,7 @@ export const App = () => {
                   <hr/>
                   <EmojiForm
                     deployedWallet={new DeployedWallet(TEST_CONTRACT_ADDRESS, 'bob.mylogin.eth', TEST_PRIVATE_KEY, sdk)}
+                    onConnectionSuccess={() => { console.log('connect'); }}
```
marekkirejczyk: Hmmm

xitronix: Its just info for developers, who will test it in @react playground. Should I do some modal for this purpose?

PR #1161, opened 2019-09-30
GasModes: Normalized all amounts to BigNumber (n*10e18)

xitronix:


marekkirejczyk:
Will get back to it a bit later.

PR #1165, opened 2019-10-01
Feature flags

alexcherman:


universal-login-commons/lib/core/utils/stringToEnumKey.ts
```diff
@@ -0,0 +1,3 @@
+export const stringToEnumKey = (anEnum: any, keyAsString: string) => {
+  return anEnum[keyAsString as any] as any;
+};
```
marekkirejczyk: Both `stringToEnumKey` and `stringToEnumKey` can be put to one file `enums.ts`.

PR #1182, opened 2019-10-02
Add IP-Geolocation Api config to Relayer config

msieczko:
Move IPGeolocationApiConfig type to commons

Rewrite Relayer config routes E2E tests to TypeScript and add necessary getters to RelayerUnderTest

Add IP-Geolocation Api config to Relayer public config and make config endpoint include it in its response

Add ipGeolocationApi config to getRelayerConfig method in startDevelopment script


marekkirejczyk:
Looks good, need to fix ci/master before merge. @JustynaBroniszewska @msieczko

PR #1238, opened 2019-10-07
Rename MessageState to MineableState

rzadp:
- It will be re-used for both messages and deployments
- Renamed `setMessageState` to more generic `setState`


universal-login-commons/lib/core/models/message.ts
```diff
@@ -35,14 +35,14 @@ export type MessageStatus = {
...
-  state: MessageState
+  state: RepositoryItemState
 };
 
 export type MessageWithFrom = PartialRequired<SignedMessage, 'from'>;
 
 export type MessageWithoutFrom = Omit<SignedMessage, 'from'>;
 
-export type MessageState = 'AwaitSignature' | 'Queued' | 'Pending' | 'Error' | 'Success';
+export type RepositoryItemState = 'AwaitSignature' | 'Queued' | 'Pending' | 'Error' | 'Success';
```
marekkirejczyk: I am not sure about the nameing and type.
I would propose sth in this direction:
```ts
export type MinableState = MessageState | DeploymentState
```

rzadp: Done.

PR #1244, opened 2019-10-07
Introduce Mineable

rzadp:


universal-login-relayer/lib/core/models/RepositoryItem.ts
```diff
@@ -0,0 +1,7 @@
+import {MessageState} from '@universal-login/commons';
+
+export interface RepositoryItem {
```
marekkirejczyk: The naming is tricky. `RepositoryItem` talks about role in the architecture, but not in domain, which is a bad naming. 
Should be something like `Minable` or `Exectuable`. I like `Mineable`

marekkirejczyk: For Type Safety I would do sth like:
```ts
type Minable = Message | Deployment;
```

rzadp: Renamed `RepositoryItem` to `Mineable`

PR #1271, opened 2019-10-08
Remove tx hash in onboarding

alexcherman:


marekkirejczyk:
Let's leave it for now. We will implement this in upcoming days.

PR #1274, opened 2019-10-09
Handle ua-parser-js types (update yarn.lock)

xitronix:


xitronix:
@Marik-D Can you check if it work for your app? It is work for Jarvis

universal-login-relayer/test/unit/http/utils/getDeviceInfo.ts
```diff
@@ -0,0 +1,113 @@
+import {expect} from 'chai';
+import {getDeviceInfo} from '../../../../lib/http/utils/getDeviceInfo';
+import {TEST_APPLICATION_INFO} from '@universal-login/commons';
+
+describe('UNIT: getDeviceInfo', () => {
+  const req = {
```
Marik-D: Can you inline it for the two tests that need it?

PR #1278, opened 2019-10-09
Relayer - extract function refactor

rzadp:


universal-login-relayer/lib/core/services/messages/ExecutionWorker.ts
```diff
@@ -28,29 +36,21 @@ class ExecutionWorker {
...
-    };
-
-    const tryExecute = async (nextMessage: QueueItem) => {
-      if (this.messageExecutor.canExecute(nextMessage)){
-        await this.execute(nextMessage.hash);
-      } else {
-        await this.executionQueue.remove(nextMessage.hash);
-      }
-    };
+  async tick() {
```
JustynaBroniszewska: make it private

rzadp: Done.

PR #1285, opened 2019-10-10
Wallet: Add waitingFor

cehali:
# Summary
Replace in Wallet ModalWaitingFor with WaitingFor from React.

Fixe: 0

## Checklist


universal-login-wallet/src/ui/react/common/ImageWaitingFor.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export const ImageWaitingFor = () => {
+  return (
+    <img
+      className="modal-avatar-pending"
+      src={AvatarPending1x}
+      srcSet={AvatarPending2x}
+      alt="pending"
+    />
+  );
```
JustynaBroniszewska: you can do it simpler

universal-login-wallet/src/core/entities/WalletModalContext.ts
```diff
@@ -11,6 +10,6 @@ export interface TopUpModalProps {
 
 type ErrorMessageType = string;
 
-export type WalletModalPropType = Partial<ModalWaitingForProps> | TopUpModalProps | ErrorMessageType;
+export type WalletModalPropType = Partial<WaitingForProps> | TopUpModalProps | ErrorMessageType;
```
JustynaBroniszewska: Why Partial?

universal-login-react/src/ui/styles/waitingFor.sass
```diff
@@ -1,64 +1,65 @@
 .universal-login-waiting-for
-  max-width: 570px
-  margin: 0 auto
-  padding-top: 120px
-  min-height: 600px
+  .universal-login-default
```
JustynaBroniszewska: I don't think that all the styles should be in `.universal-login-defaults`. Colors should be defaults for sure, but height, width, etc. should be in `.universal-login-waiting-for` class, not only in `.universal-login-default` class. 

The way you did it `.universal-login-waiting-for` class seems to be redundant.

universal-login-wallet/src/core/entities/WalletModalContext.ts
cehali: Because not all values are set at the same moment.

universal-login-wallet/src/ui/react/Modals/Modal.tsx
```diff
@@ -42,15 +42,28 @@ const Modal = () => {
...
     case 'waitingForDeploy':
-      return (
+    return (
         <ModalWrapper modalClassName="jarvis-modal">
-          <ModalWaitingFor {...modalService.modalProps} action={'Wallet creation'} chainName={relayerConfig!.chainSpec.name} transactionHash={'0xee9270ccdeb9fcb92b3ec509ba11ba2362ab32ba8f...'}/>
+          <WaitingFor
+            {...modalService.modalProps}
+            action={'Wallet creation'}
+            relayerConfig={relayerConfig!}
+            transactionHash={'0xee9270ccdeb9fcb92b3ec509ba11ba2362ab32ba8f...'}
```
marekkirejczyk: We can now implement a follow-up ticket and put real hash here.
Next ticket?

PR #1291, opened 2019-10-11
Relayer: Export MinedTransactionHandler

xitronix:


marekkirejczyk:
Tests should be refactored so that MinedMessageHandler has his own test suite

PR #1300, opened 2019-10-11
[Optional] React: Screenshots

rzadp:
Optional feature - making screenshots for React app.
The screenshots are available in [artifacts](https://circleci.com/gh/UniversalLogin/UniversalLoginSDK/23908#artifacts/containers/0) page:

Screenshot 2019-10-11 at 23 41 51

### puppeteer
The `puppeteer` is sometimes slow to install, but is a devDependency, so it should be OK as long as we change `yarn` to `yarn install --prod` on heroku, netlify, etc.

### CI
In order to not impact CI time, it is only run on `master` and on branches starting with `ss-`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,8 @@ workflows:
       - test_sdk_relayer
       - test_react_wallet_web3
       - react_screenshots:
-          branches:
-            only:
-              - master
-              - /ss-.*/
+          filters:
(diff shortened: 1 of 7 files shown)


marekkirejczyk:
Awesome idea. Merging to play with it. If helpful would be good to refactor scrennshots.js.
Also could be do for wallet and on all branches.

PR #1321, opened 2019-10-16
Load relayer config in wallet

rzadp:
Fix analogical to #1318


Marik-D:
Shouldn't it be already loaded when sdk is started?

rzadp:
I dont’t know, but it was failing for me when running the wallet locally
using start:dev

Marik-D:
Yes, I had this problem too. The way I fixed it was to wait for `sdk.start` to complete before rendering the app. https://github.com/UniversalLogin/UniversalLoginSDK/pull/1320/files#diff-21fe2ad581fd37458813f15f961f1d8aR11

Marik-D:
Also, since we are fetching config in `sdk.start`, maybe we should make `sdk.fetchRelayerConfig` private to prevent re-fetching

PR #1325, opened 2019-10-16
Rename WaitingFor to WaitingForTransaction

JustynaBroniszewska:


universal-login-react/src/ui/commons/WaitingFor.tsx
```diff
@@ -34,11 +34,11 @@ const renderWaitingFor = (action: string, relayerConfig: PublicRelayerConfig, tr
   );
 };
 
-export const WaitingFor = ({action, relayerConfig, transactionHash, children, className}: WaitingForProps) => {
+export const WaitingForTransaction = ({action, relayerConfig, transactionHash, children, className}: WaitingForTransactionProps) => {
```
rzadp: Rename file name to WaitingForTransaction.tsx ?

cehali: 👍

JustynaBroniszewska: I will extract WaitingFor and this component will be in this file, so that is why I didn't rename the file name.

JustynaBroniszewska: @rzadp

PR #1343, opened 2019-10-17
Remove setBetaNotice() duplication

alexcherman:
We already setting notice message here: https://github.com/UniversalLogin/UniversalLoginSDK/blob/master/universal-login-wallet/src/index.tsx


JustynaBroniszewska:
Resolve conflicts

PR #1346, opened 2019-10-18
Enable pay without amount

xitronix:


xitronix:
I change it in another PR

PR #1347, opened 2019-10-18
Introduce UNIVERSAL_LOGIN_LOGO_URL

JustynaBroniszewska:


universal-login-react/src/ui/UFlow/Devices/Logo.tsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react';
+import {UNIVERSAL_LOGIN_LOGO_URL as DEFAULT_LOGO} from '@universal-login/commons';
```
xitronix: I would prefer 
```ts
const DEFAULT_LOGO  = UNIVERSAL_LOGO_URL
```
Because it default logo has another meaning and related context or like Justyna guess in the start just leave ```ts
import {UNIVERSAL_LOGIN_LOGO_URL} ..
```
to be sure that if we change universal logo, default logo change too.
@JustynaBroniszewska, @rzadp, @cehali  what do you think?

PR #1356, opened 2019-10-18
Absolute path for manage devices

rzadp:


rzadp:
Does not work in jarvis

PR #1360, opened 2019-10-18
Show waitingForOnRampProvider after providers modal close

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/TopUp.tsx
```diff
@@ -49,6 +52,10 @@ export const TopUp = ({sdk, onGasParametersChanged, contractAddress, startModal,
     />
   );
 
+  const showWaitingFor = (provider: string) => {
```
xitronix: Try to change string for proper type

PR #1363, opened 2019-10-21
Use Router absolute path in Devices

rzadp:


rzadp:
Merge after https://github.com/UniversalLogin/UniversalLoginSDK/pull/1324

PR #1368, opened 2019-10-21
Delete account in web3

Marik-D:
# Summary

Fixes #number

## Checklist


universal-login-react/src/App.tsx
```diff
@@ -59,7 +60,7 @@ export const App = () => {
     return () => {};
   };
 
-  const [deployedWallet, setDeployedWallet] = useState(new DeployedWallet('', '', '', sdk));
+  const [walletService] = useState(new WalletService(sdk));
```
rzadp: Why not just `const walletService = new WalletService(sdk)` ?

universal-login-react/src/ui/UFlow/DashboardModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const {relayerConfig} = walletService.sdk;
+
+  const deployedWallet = walletService.getDeployedWallet();
+  const {contractAddress, sdk, name} = deployedWallet;
+
+  const [transferDetails, setTransferDetails] = useState<Partial<TransferDetails>>({transferToken: sdk.tokensDetailsStore.tokensDetails[0].address} as TransferDetails);
+  const selectedToken = sdk.tokensDetailsStore.getTokenByAddress(transferDetails.transferToken!);
+  const [transactionHash, setTransactionHash] = useState('');
+
+  const notice = sdk.getNotice();
```
cehali: This shouldn't be in useEffect?

Marik-D: No, `getNotice` just reads a variable so no need for `useEffect` here

universal-login-react/src/App.tsx
cehali: 👍

universal-login-react/src/ui/UFlow/LogoButton.tsx
```diff
@@ -1,21 +1,15 @@
...
-  if (deployedWallet.contractAddress.length > 10) {
-    return (
-      <div className={getStyleForTopLevelComponent(className)}>
-        <Dashboard deployedWallet={deployedWallet} />
-      </div>
-    );
-  } else {
-    return <></>;
-  }
+export const LogoButton = ({className, ...rest}: LogoButtonProps) => {
```
cehali: restProps? or just props?

JustynaBroniszewska:
Needs rebase

universal-login-react/src/App.tsx
Marik-D: > Why not just const walletService = new WalletService(sdk) ?

That way, for each re-render new `WalletService` would be created

Marik-D:
@JustynaBroniszewska @rzadp @cehali Please re-review after rebase

universal-login-react/src/App.tsx
rzadp: Are you sure that with `useState`, new `WalletService` will be created more rarely?

I am new to functional components and hooks so I don't know, but quick experiment with `console.log` in `WalletService`'s constructor shows that new instance is created with `useState` as well.

Marik-D:
@JustynaBroniszewska @rzadp @cehali Please re-review after another rebase

universal-login-react/src/App.tsx
Marik-D: hmm, thats odd, could you send me an example?

Marik-D: On a second thought: the code above will actually create a new wallet service every render, but that wallet service is **discarded** on subsequent renders

rzadp: I've approved and it is not important, but I think the conversation is interesting. What do you mean "discarded"? Are you talking about garbage collection?

Marik-D: @rzadp 
Lets compare the two pieces of code (assume its inside a render method):
```
// 1
const walletService = new WalletService()

// 2
const [walletService] = useState(new WalletService())
```

Although in both example a new instance of `WalletService` is created each render, the variable  `walletService`  will always hold the same instance in the second example, while containing a new value every render in a first example.

There is also a special syntax to avoid needlessly creating a new instance of `WalletService` every render:

```
const [walletService] = useState(() => new WalletService())
```

React will call the lambda only once: on the initial render.

rzadp: OK I think I get it, thanks.

PR #1376, opened 2019-10-22
Increase deploy gas limit

rzadp:
Currently the deployment fails on Kovan with `out of gas`


xitronix:
@JustynaBroniszewska @rzadp I think we need to check how much deploying cost and increase refund in contracts and constants too

JustynaBroniszewska:
Does it work on other testnets?

rzadp:
> Does it work on other testnets?

I just checked, works for:
- Rinkeby
- Ropsten
- and Kovan of course

PR #1392, opened 2019-10-23
Resolve bug when canceling connection from connecting device

xitronix:


JustynaBroniszewska:
We will do it in the different way

PR #1393, opened 2019-10-23
Single restricted pattern

rzadp:


.eslintrc.json
```diff
@@ -331,10 +331,9 @@
     "no-unsafe-negation": "error",
     "no-restricted-imports": ["error", {
       "patterns": [
-        "@universal-login/*/src",
-        "@universal-login/*/lib",
-        "@universal-login/*/test",
-        "@universal-login/*/dist"
+        "@universal-login/*/*",
+        "!@universal-login/*/testutils",
+        "!@universal-login/contracts/*"
```
JustynaBroniszewska: `"!@universal-login/contracts/build"` we don't want to import from `"!@universal-login/contracts/lib"`

rzadp: Done

PR #1406, opened 2019-10-23
EmojiForm and ConnectionNotification improvement

JustynaBroniszewska:


universal-login-react/src/App.tsx
```diff
@@ -120,16 +120,19 @@ export const App = () => {
...
-                <div>
-                  <EmojiPanel code={generateCode(CONNECTION_REAL_ADDRESS)} />
-                  <hr />
-                  <EmojiForm
-                    deployedWallet={new DeployedWallet(TEST_CONTRACT_ADDRESS, 'bob.mylogin.eth', TEST_PRIVATE_KEY, sdk)}
-                    onConnectionSuccess={() => {console.log('connect');}}
-                  />
-                </div>
-              )}
+              render={() => {
```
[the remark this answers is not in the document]
Marik-D: IMHO looks better than `render`

universal-login-react/src/ui/Notifications/EmojiForm.tsx
```diff
@@ -4,45 +4,32 @@ import {
...
-  const [notifications, setNotifications] = useState<Notification[]>([]);
   const [soleAddress, setSoleAddress] = useState<string | undefined>(undefined);
 
-  useEffect(() => deployedWallet.subscribeAuthorisations(onNotificationsChange), []);
-
-  const onNotificationsChange = (notifications: Notification[]) => {
-    setNotifications(notifications);
-    updateSoleAddress(filterNotificationByCodePrefix(notifications, enteredCode));
-  };
+  useEffect(() => updateSoleAddress(filterNotificationByCodePrefix(notifications, enteredCode)), [notifications]);
```
Marik-D: You can remove the call to `updateSoleAddress` in `updateEnteredCode` and just add `enteredCode` as a dependency to this effect.

universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
```diff
@@ -17,38 +22,81 @@ interface ConnectNotificationProps {
...
 
-  const onDenyButtonClick = () => {
-    deployedWallet.denyRequests();
+  const onDenyButtonClick = async () => {
+    await deployedWallet.denyRequests();
     history.goBack();
   };
 
+  const onCancelClick = async () => {
+    await deployedWallet.denyRequests();
```
Marik-D: Look really similar to `onDenyButtonClick`. Maybe this one shouldn't deny requests

universal-login-react/src/App.tsx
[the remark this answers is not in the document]
JustynaBroniszewska: Ok, but look at the whole file. We render all components like that. Let's do it in another PR #makePlaygroundAppGreatAgain

universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
rzadp: Is seems to be refered as a "reject" button:
https://github.com/UniversalLogin/UniversalLoginSDK/blob/9dc642b18798d6c4e54ca00f8e7b94dd1e78ca38/universal-login-react/src/ui/Notifications/EmojiForm.tsx#L66-L70

Shall we rename this to `onRejectClick`? Also then it would make sense that "reject" denies the requests

PR #1412, opened 2019-10-25
React: Enable missing amount if provider does not use amount

rzadp:


universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -93,7 +93,10 @@ export const TopUpWithFiat = ({sdk, onPayClick, logoColor}: TopUpWithFiatProps)
         <button
           onClick={() => onPayClick(paymentMethod!, amount)}
           className="pay-btn"
-          disabled={!paymentMethod || !amount}
+          disabled={
+            !paymentMethod ||
+            (topUpProviderSupportService.isInputAmountUsed(paymentMethod) && !amount)
```
cehali: Maybe extract this condition to separate function

rzadp: Done.

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -56,6 +56,10 @@ export const TopUpWithFiat = ({sdk, onPayClick, logoColor}: TopUpWithFiatProps)
 
   useAsyncEffect(recognizeUserCountry, []);
 
+  const isPayButtonDisabled = () =>
```
Marik-D: What's the reasoning behind using a function here instead of just a plain constant?

PR #1418, opened 2019-10-25
Show minimal-amount

xitronix:


universal-login-react/src/ui/TopUp/ChooseTopUpMethod.tsx
```diff
@@ -25,11 +25,19 @@ export interface ChooseTopUpMethodProps {
...
   }
+  const gasParametersChanged = (gasParameters: GasParameters) => {
+    if (isDeployment) {
+      setGasParameters(gasParameters);
+      onGasParametersChanged!(gasParameters);
+    }
+  };
   const [topUpMethod, setTopUpMethod] = useState('');
   const methodSelectedClassName = topUpMethod !== '' ? 'method-selected' : '';
+  const getMinimalAmount = (gasParameters?: GasParameters) => gasParameters && safeMultiply(DEPLOYMENT_REFUND.mul(3).div(2), gasParameters.gasPrice);
```
JustynaBroniszewska: What is this?  😨

xitronix: As @marekkirejczyk said we just show 150% of deployment fee as a minimal amount. So I can rename and move this function

universal-login-react/src/ui/TopUp/ChooseTopUpMethod.tsx
```diff
@@ -25,11 +25,19 @@ export interface ChooseTopUpMethodProps {
...
+  const [gasParameters, setGasParameters] = useState<GasParameters | undefined>(undefined);
   if (isDeployment) {
     ensureNotNull(onGasParametersChanged, MissingParameter, 'onGasParametersChanged');
   }
+  const gasParametersChanged = (gasParameters: GasParameters) => {
+    if (isDeployment) {
+      setGasParameters(gasParameters);
+      onGasParametersChanged!(gasParameters);
+    }
+  };
```
cehali: Do we need this check if `gasParametersChanged` is used only in GasPrice component which is shown only if isDeployment === true

PR #1447, opened 2019-10-29
Improve hint styling

xitronix:
image

Screenshot 2019-10-29 at 12 14 50


universal-login-react/src/ui/styles/hint.css
```diff
@@ -1,8 +1,9 @@
 .hint {
-  margin-top: 20px;
+  margin-top: 1px;
```
alexcherman: Probably unnecessary

xitronix: ✔️

PR #1455, opened 2019-10-29
Styling disabled send btn when gas price loading

xitronix:
Screenshot 2019-10-29 at 16 32 38
Screenshot


rzadp:
Found some problems in Jarvis, probably not for this PR:

In Jarvis, when gas price is still loading, clicking Crypto breaks the app

Screenshot 2019-10-29 at 17 10 34

Screenshot 2019-10-29 at 17 10 38

PR #1458, opened 2019-10-29
Go back if notifications become empty

xitronix:


universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
```diff
@@ -50,23 +55,19 @@ export const ConnectionNotification = ({deployedWallet, devicesBasePath, classNa
     <div id="notifications" className="universal-login-emojis">
       <div className={getStyleForTopLevelComponent(className)}>
         <div className="approve-device">
-          {notifications.length > 0 && (
+          {showTitle &&
```
cehali: Maybe we should think about changing name because this variable defines if we show the whole text not the only title.

universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
```diff
@@ -50,23 +55,19 @@ export const ConnectionNotification = ({deployedWallet, devicesBasePath, classNa
...
-                onCancelClick={onCancelClick}
-                setPublicKey={setPublicKey}
-              />
+              <p className="approve-device-title">Approve device</p>
+              <p className="approve-device-text">A new device tries to connect to this account. Enter emojis in the correct order to approve it.</p>
             </>
-          )}
+          }
+          <EmojiForm
+            hideTitle={() => setShowTitle(false)}
```
cehali: The same as above.

universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
rzadp: `showHeader` maybe, the header would be title+text.
I'd say this rename is optional in this PR.

universal-login-react/src/ui/Notifications/ConnectionNotification.tsx
```diff
@@ -22,10 +22,15 @@ export const ConnectionNotification = ({deployedWallet, devicesBasePath, classNa
...
   const history = useHistory();
 
+  const updateNotifications = (notifications: Notification[]) => notifications.length === 0
+    ? history.goBack()
+    : setNotifications(notifications);
+
+  useEffect(() => deployedWallet.subscribeAuthorisations(updateNotifications), []);
+
   const onDenyClick = async () => {
     await deployedWallet.denyRequests();
```
rzadp: Is `updateNotifications` callback executed after `await deployedWallet.denyRequests()`?
If so, then `history.goBack()` happens, and then we have `history.goBack()` again in the next line?

xitronix: Doesn't brake, but it is better for user experience. Done 👍

PR #1460, opened 2019-10-30
Return waitToBeSuccess and waitForTransactionHash in generateBackupCodes

JustynaBroniszewska:
# Summary
Return execution and codes in generateBackupCodes to be able to use waitForTransactionHash in React

Fixes: 0

## Checklist

### DIFF
--- a/universal-login-sdk/lib/api/DeployedWallet.ts
+++ b/universal-login-sdk/lib/api/DeployedWallet.ts
@@ -108,10 +108,10 @@ export class DeployedWallet implements ApplicationWallet {
     const execution = await this.sdk.addKeys(this.contractAddress, addresses, this.privateKey, {gasToken: ETHER_NATIVE_TOKEN.address, gasPrice: DEFAULT_GAS_PRICE, gasLimit: DEFAULT_GAS_LIMIT});
     return {
       waitToBeSuccess: async () => {
-        await execution.waitToBeSuccess()
+        await execution.waitToBeSuccess();
         return codes;
       },
-      waitForTransactionHash: execution.waitForTransactionHash
+      waitForTransactionHash: execution.waitForTransactionHash,
     };
   }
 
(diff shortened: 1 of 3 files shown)


universal-login-sdk/lib/api/DeployedWallet.ts
```diff
@@ -99,8 +104,7 @@ export class DeployedWallet implements ApplicationWallet {
     }
 
     const execution = await this.sdk.addKeys(this.contractAddress, addresses, this.privateKey, {gasToken: ETHER_NATIVE_TOKEN.address, gasPrice: DEFAULT_GAS_PRICE, gasLimit: DEFAULT_GAS_LIMIT});
-    await execution.waitToBeSuccess();
-    return codes;
+    return {codes, execution};
```
rzadp: How about:
```
return {
    waitToBeSuccess: async () => {
        await execution.waitToBeSuccess();
        return codes;
  }
};
```

Marik-D: 👍

Marik-D: That would mean that we'd need to always wait for execution to succeed to get the backup codes

JustynaBroniszewska: And that is correct because we don't want to have codes that are not in the contract.

universal-login-sdk/lib/api/DeployedWallet.ts
```diff
@@ -15,6 +16,12 @@ import WalletContract from '@universal-login/contracts/build/Wallet.json';
 import {BigNumber} from 'ethers/utils';
 import {OnBalanceChange} from '../core/observers/BalanceObserver';
 
+interface BackupCodesWithExecution {
+  waitToBeSuccess: () => Promise<string[]>;
+  waitForTransactionHash: () => Promise<MessageStatus>;
+
```
rzadp:
```suggestion
```

PR #1462, opened 2019-10-30
Add HTTPS redirect to the relayer

Marik-D:
Set `HTTPS_REDIRECT=true` env variable to enable the redirect


universal-login-relayer/lib/http/relayers/Relayer.ts
```diff
@@ -98,11 +99,17 @@ class Relayer {
 
   runServer() {
     this.app = express();
+    this.app.set('trust proxy', true);
```
rzadp: Why?

Marik-D: so the `req.protocol` is assigned from `X-Forwarded-Proto`

PR #1472, opened 2019-10-30
Add privacy policy

JustynaBroniszewska:


JustynaBroniszewska:
Small change: 
Screenshot 2019-10-30 at 16 28 07

universal-login-wallet/src/ui/react/Home/PrivacyPolicy.tsx
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable */
+import React from 'react';
+
+export const PrivacyPolicy = () => {
+  return (
+    <div className="privacy-description">
```
PetarKirov: When we start working on multi language support, perhaps it would be better if we use markdown files for each document like that.

PR #1474, opened 2019-10-30
Change push to replace

cehali:
# Summary
Change push to replace in Devices and Backup codes


JustynaBroniszewska:
I am not sure why we do this?

PR #1482, opened 2019-10-31
Agree button style

rzadp:


rzadp:
Before:

Screenshot 2019-10-31 at 11 28 17

rzadp:
After:

Screenshot 2019-10-31 at 11 31 22

PR #1487, opened 2019-10-31
Top up refactoring

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/ChooseTopUpMethod.tsx
```diff
@@ -54,82 +50,53 @@ export const ChooseTopUpMethod = ({sdk, contractAddress, onPayClick, topUpClassN
...
-                  <img className="top-up-method-icon" src={bankIcon} alt="Dai" />
-                </div>
-                <p className="top-up-method-title">Fiat</p>
-                <p className="top-up-method-text">Buy using credit card or bank account</p>
-              </TopUpRadio>
-
-            </div>
-          </div>
-          <div className="top-up-body">
-            <div className="top-up-body-inner">
```
xitronix: I think we should keep it's here

PR #1491, opened 2019-10-31
Wallet: Replace info text after wallet created

cehali:
# Summary
Replace info text after wallet created in ModalTxnSuccess


cehali:
Before: 
image

cehali:
After:
image

PR #1500, opened 2019-11-04
Fix max length warnigns

xitronix:


universal-login-contracts/lib/encode.ts
```diff
@@ -42,6 +42,7 @@ export function setupInitializeWithENSArgs({keyPair, ensDomainData, gasPrice, ga
   const ensName = `${name}.${domain}`;
   const hashLabel = utils.keccak256(utils.toUtf8Bytes(name));
   const node = utils.namehash(ensName);
-  const args = [keyPair.publicKey, hashLabel, ensName, node, ensDomainData.ensAddress, ensDomainData.registrarAddress, ensDomainData.resolverAddress, gasPrice, gasToken];
+  const args =
```
JustynaBroniszewska: Inline

xitronix: changed for: 
```ts
  const args = [
    keyPair.publicKey,
    hashLabel,
    ensName,
    node,
    ensDomainData.ensAddress,
    ensDomainData.registrarAddress,
    ensDomainData.resolverAddress,
    gasPrice,
    gasToken,
  ];
```

rzadp:
> Why not just remove this rule? It would be better and easier.

Agree - let's remove the rule

xitronix:
If we have legth of line > 160, it often means that code is unreadable.

xitronix:
So

PR #1501, opened 2019-11-04
Wallet: Delete top padding in new device

cehali:


cehali:
Before:
image
image

cehali:
After:
image
image

cehali:
How it looks on mobile: 
image

PR #1515, opened 2019-11-04
Store Future wallet in local storage

Marik-D:
# Summary

Fixes #number

## Checklist


universal-login-react/src/core/services/WalletStorageService.ts
```diff
@@ -1,34 +1,65 @@
 import {StorageEntry, StorageService} from './StorageService';
-import {ApplicationWallet} from '@universal-login/commons';
-import {asObject, asString, Sanitizer} from '@restless/sanitizers';
-import {WalletStorage} from '@universal-login/sdk';
+import {ApplicationWallet, asExactly, CounterfactualWallet} from '@universal-login/commons';
+import {asAnyOf, asObject, asString, cast} from '@restless/sanitizers';
+import {WalletStorage, SerializedWalletState} from '@universal-login/sdk';
+
+const STORAGE_KEY = 'wallet';
```
marekkirejczyk: Should we use more sophisticated storage key, e.g. `wallet.jarvis.network`?

universal-login-react/test/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,70 @@
+import {StorageService, WalletStorageService} from '../../../src';
+import {ApplicationWallet, CounterfactualWallet} from '@universal-login/commons';
+import sinon from 'sinon';
+import {expect} from 'chai';
+
+describe('WalletStorageService', () => {
+  const counterfactualWallet: CounterfactualWallet = {
+    contractAddress: '0xFOO',
```
marekkirejczyk: A good practice is to use real parsable addresses.

universal-login-react/test/core/services/WalletStorageService.ts
```diff
@@ -0,0 +1,70 @@
+import {StorageService, WalletStorageService} from '../../../src';
+import {ApplicationWallet, CounterfactualWallet} from '@universal-login/commons';
+import sinon from 'sinon';
+import {expect} from 'chai';
+
+describe('WalletStorageService', () => {
```
marekkirejczyk: Not sure if those tests test a lot. I would consider writing MemoryStorageService, so we can write round trips tests.

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -22,11 +29,19 @@ type WalletState = {
   wallet: DeployedWallet;
 };
 
+export type SerializedWalletState = {
+  kind: 'Future';
+  wallet: CounterfactualWallet;
+} | {
+  kind: 'Deployed';
+  wallet: ApplicationWallet;
+};
```
marekkirejczyk: That looks like duplication topped with some extra name mixing.

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -163,29 +179,53 @@ export class WalletService {
...
   disconnect(): void {
     this.stateProperty.set({kind: 'None'});
-    this.removeFromStorage();
+    this.saveToStorage();
   }
 
-  saveToStorage(applicationWallet: ApplicationWallet) {
-    this.storage && this.storage.save(applicationWallet);
+  saveToStorage() {
+    if (!this.storage) {
```
marekkirejczyk: Why not throw an exception? What would be the use-case for WalletService without storage?

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -163,29 +179,53 @@ export class WalletService {
...
+        kind: 'Future',
+        wallet: await this.sdk.getFutureWalletFactory().createFromExistingCounterfactual(state.wallet),
+      });
+    } else if (state.kind === 'Deployed') {
+      this.stateProperty.set({
+        kind: 'Deployed',
+        wallet: new DeployedWallet(state.wallet.contractAddress, state.wallet.name, state.wallet.privateKey, this.sdk),
+      });
+    } else {
+      console.error('Invalid saved wallet state', state);
```
marekkirejczyk: Why not throw an exception? Fail fast.

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -163,29 +179,53 @@ export class WalletService {
...
+          contractAddress: this.state.wallet.contractAddress,
+          privateKey: this.state.wallet.privateKey,
+        },
+      });
+    } else if (this.state.kind === 'Deployed') {
+      this.storage.save({
+        kind: 'Deployed',
+        wallet: this.state.wallet.asApplicationWallet,
+      });
+    }
```
marekkirejczyk: Could be:
```ts
saveToStorage() {
  this.storage.save(this.serializer.serialize(this.state));
}
```

universal-login-sdk/lib/core/services/WalletService.ts
```diff
@@ -163,29 +179,53 @@ export class WalletService {
...
   }
 
-  loadFromStorage() {
+  async loadFromStorage() {
```
marekkirejczyk: Could be:
```ts
async loadFromStorage() {
  this.state = this.serializer.deserialize(this.storage.load());
}
```

universal-login-react/src/core/services/WalletStorageService.ts
Marik-D: This is a good idea, but Jarvis already uses the `wallet` key for storage. We might want to consider some more sophisticated migration mechanisms if we decide to change the key or add versioning

universal-login-sdk/lib/core/services/WalletService.ts
Marik-D: > Careful with naming. Let's be consistent here. What is the difference between CounterfactualWallet and FutureWallet? Perhaps you meant SerializableFutureWallet?

Good idea should we also consider renaming `ApplicationWallet` to `SerializableDeployedWallet`?

> We should probably extract class WalletSerializer that converts WalletState to SerializableWalletState, with two methods:

The `WalletService.loadFromStorage` method will soon contain side effects in addition to deserialising. This doesn't fit well with serializer pattern.

universal-login-react/test/core/services/WalletStorageService.ts
Marik-D: 👍

universal-login-react/test/core/services/WalletStorageService.ts
Marik-D: 👍

universal-login-sdk/lib/core/services/WalletService.ts
Marik-D: 👍

universal-login-sdk/lib/core/services/WalletService.ts
Marik-D: 👍

universal-login-sdk/lib/core/services/WalletService.ts
Marik-D: In-memory wallet only. We do this in playground for example

universal-login-sdk/lib/core/services/WalletService.ts
Marik-D: > That could be split into a number of files:

Separated the models to a separate file

universal-login-react/src/core/services/WalletStorageService.ts
Marik-D: https://github.com/EthWorks/original-works/tree/master/original-works-mobile/src/services/migrations

rzadp:
I seem to have these errors on the console when running react:

Screenshot 2019-11-05 at 11 02 49

Can a third person try to run it and check it?

rzadp:
BTW I have these erros in the console as well:
Screenshot 2019-11-05 at 11 13 35

rzadp:
Should we change the way this is saved to local storage to keep it consistent?

https://github.com/UniversalLogin/UniversalLoginSDK/blob/590668103fe6956dc01c75032d5be7adab898570/universal-login-react/src/ui/Onboarding/Onboarding.tsx#L50-L54

universal-login-react/src/core/services/WalletStorageService.ts
marekkirejczyk: Exactly, but with YAGNI, I would start with a very simple migration mechanism (just one method) and evolve it into a full-blown migration mechanism if needed. 
It can be done in future PRs.

universal-login-sdk/lib/core/services/WalletService.ts
marekkirejczyk: Shall we then implement an in-memory store as a separate class (for test and playground)?

Marik-D: `WalletService` already has it's own internal in-memory state storage. We cannot fully offload the state to injectable storage since it has some unserializable parts. So creating a separate `MemoryWalletStorageService` right now isn't needed since it would have the same data lifetime as `WalletService`'s internal state store.

xitronix:
I don't have any errors in console log when i run react

universal-login-sdk/lib/core/models/WalletService.ts
```diff
@@ -0,0 +1,32 @@
...
+  wallet: ApplicationWallet;
+} | {
+  kind: 'Deploying';
+  wallet: ApplicationWallet;
+} | {
+  kind: 'Deployed';
+  wallet: DeployedWallet;
+};
+
+export type SerializedWalletState = {
```
marekkirejczyk: Shall we extract SerializedWalletState to a separate file?

universal-login-sdk/lib/api/FutureWalletFactory.ts
```diff
@@ -24,7 +24,7 @@ export interface Deployment {
   waitToBeSuccess: () => Promise<DeployedWallet>;
 }
 
-export interface FutureWallet extends CounterfactualWallet {
+export interface FutureWallet extends SerializableFutureWallet {
```
marekkirejczyk: Why do we need to extend?

Marik-D: We don't, but `FutureWallet` has the same data fields as `SerializableFutureWallet`. `extend` is an alternative to just repeating those field in there. Do you think it would be better to repeat them instead?

universal-login-sdk/lib/core/services/WalletSerializer.ts
```diff
@@ -0,0 +1,46 @@
+import {SerializedWalletState, WalletState} from '../models/WalletService';
+import UniversalLoginSDK, {DeployedWallet} from '../..';
+
+export class WalletSerializer {
+  constructor(
+    private readonly sdk: UniversalLoginSDK,
+  ) {
+  }
+
+  serialize(state: WalletState): SerializedWalletState | 'remove' | undefined {
```
marekkirejczyk: Why not just `SerializedWalletState?`?

PR #1519, opened 2019-11-05
Display UI based on wallet state in create flow

Marik-D:


xitronix:
Let's rebase first?

universal-login-wallet/src/ui/react/CreateAccount/CreateFlow.tsx
```diff
@@ -0,0 +1,78 @@
...
+import {ModalWrapper, TopUp, useProperty, WaitingForDeployment} from '@universal-login/react';
+import {ModalTxnSuccess} from '../Modals/ModalTxnSuccess';
+
+export function CreateFlow() {
+  const modalService = useContext(WalletModalContext);
+  const {sdk, walletService} = useServices();
+  const [gasParameters, setGasParameters] = useState<GasParameters>(INITIAL_GAS_PARAMETERS);
+  const [transactionHash, setTransactionHash] = useState<string | undefined>(undefined); // TODO: Move to wallet state
+
+  const deployWallet = useMutableCallback((name: string) => walletService.deployFutureWallet(
```
Marik-D: Interesting case here. Because of the way javascript arrow functions capture variables, if I were to call `walletService.deployFutureWallet` directly in `onCreateClick` the `gasParameters` value used would be the one that was at the beginning of the async call. `useMutableCallback` here always updates the `deployWallet` function with new values each render. This way even ongoing async operation gets up-to-date data.

Marik-D: Alternative to this code would be:
```
const [getGasParameters, setGasParameters] = useMutableState<GasParameters>(INITIAL_GAS_PARAMETERS);

const onCreateClick(name: string) => {
  ...
  await walletService.deployFutureWallet(name, getGasParameters());
}
```
In this case since we capture the getter for gas parameters instead of the variable itself, the value can be kept up to date.

Marik-D: I'm leaning to changing the code to the second version, let me know what you think

universal-login-wallet/src/ui/react/CreateAccount/CreateFlow.tsx
```diff
@@ -0,0 +1,76 @@
...
+import {ETHER_NATIVE_TOKEN, GasParameters, INITIAL_GAS_PARAMETERS} from '@universal-login/commons';
+import {hideTopUpModal} from '../../../core/utils/hideTopUpModal';
+import Modal from '../Modals/Modal';
+import {ModalWrapper, TopUp, useProperty, WaitingForDeployment} from '@universal-login/react';
+import {ModalTxnSuccess} from '../Modals/ModalTxnSuccess';
+
+export function CreateFlow() {
+  const modalService = useContext(WalletModalContext);
+  const {sdk, walletService} = useServices();
+  const [getGasParameters, setGasParameters] = useMutableState<GasParameters>(INITIAL_GAS_PARAMETERS);
```
Marik-D: Interesting case here. Because of the way javascript functions capture variables, using normal `useState` here would result in `gasPrice` not being updated midway through `onCreateClick` execution. This is because with each render `useState` technically returns a new variable, and the old ones keep their old values. `useMutableState` on the other hand returns a getter. This makes it possible to make sure that call to this getter always returns the latest values that was set.

Here's the implementation of `useMutableState` for reference:
https://github.com/Marik-D/react-use-mutable/blob/master/src/useMutableState.ts

rzadp: 👍

universal-login-wallet/package.json
```diff
@@ -30,7 +30,8 @@
     "react-dom": "^16.8.6",
     "react-router": "^5.1.2",
     "react-router-dom": "^5.1.2",
-    "react-transition-group": "^2.8.0"
+    "react-transition-group": "^2.8.0",
+    "react-use-mutable": "^0.1.1"
```
marekkirejczyk: No production of micro-npms, they introduce additional dependency management and security risk.

universal-login-wallet/src/ui/react/CreateAccount/CreateAccount.tsx
```diff
@@ -1,46 +1,18 @@
...
-    showWaitingModal();
-    await walletService.deployFutureWallet(
-      name,
-      gasParameters.gasPrice.toString(),
-      ETHER_NATIVE_TOKEN.address,
-      showWaitingModal,
-    );
-    modalService.showModal('transactionSuccess');
-  };
+export const CreateAccount = ({onCreateClick}: CreateAccountProps) => {
```
marekkirejczyk: I don't think this component is used anymore. If so we should delete it.

universal-login-wallet/src/ui/react/CreateAccount/CreateFlow.tsx
marekkirejczyk: Can we use bind instead?
```ts
return <CreateAccount onCreateClick={onCreateClick.bind(this)}/>;
```

marekkirejczyk: How we can solve in a simpler way?

universal-login-wallet/src/ui/react/CreateAccount/CreateFlow.tsx
```diff
@@ -0,0 +1,76 @@
...
+      getGasParameters().gasPrice.toString(),
+      ETHER_NATIVE_TOKEN.address,
+      setTransactionHash,
+    );
+  };
+
+  const walletState = useProperty(walletService.stateProperty);
+  switch (walletState.kind) {
+    case 'None':
+      return <CreateAccount onCreateClick={onCreateClick}/>;
```
marekkirejczyk: `onCreateClick.bind(this)`

universal-login-wallet/src/ui/react/CreateAccount/CreateFlow.tsx
marekkirejczyk:
```ts
const deploymentService = new DeploymentService(walletService);
...
<... onGasPrametersChanged=deploymentService.setGasParameters() ...>
...
<... onCreateCreateClick={deploymentService.deploy} ...>
```

universal-login-wallet/src/core/app/WalletCreationService.ts
```diff
@@ -0,0 +1,35 @@
+import {WalletService} from '@universal-login/sdk';
+import {GasParameters, INITIAL_GAS_PARAMETERS} from '@universal-login/commons';
+import {State} from 'reactive-properties';
+
+export class WalletCreationService {
```
JustynaBroniszewska: Let's move it to react

Marik-D: #nextpr

cehali: 👍

PR #1522, opened 2019-11-06
Prepare for move EnoghTokenValidator to commons

xitronix:
Moved errors, interface and resolve depend on contracts


xitronix:
Split to 2 prs

PR #1525, opened 2019-11-06
Wallet: Change style of delete account button

cehali:
# Summary
Change style of delete account button


cehali:
Before:
image

cehali:
After:
image

rzadp:
Mobile screenshots please?

JustynaBroniszewska:
I think that if there is a lot of existing devices we should make the content scrollable. The screen shouldn't be higher than desktop.

rzadp:
> I think that if there is a lot of existing devices we should make the content scrollable. The screen shouldn't be higher than desktop.

Agree, max-width with overflow: scrollable or something

cehali:
@JustynaBroniszewska @rzadp I implemented your propositions

cehali:
image

cehali:
image

cehali:
image

cehali:
image

cehali:
The newest version:
image
image
image
image
image
image
image

PR #1536, opened 2019-11-07
Validate ens name before deploying

rzadp:


universal-login-sdk/lib/core/utils/errors.ts
```diff
@@ -69,7 +69,7 @@ export class ValidationFailed extends SDKError {
 
 export class InvalidAddressOrEnsName extends ValidationFailed {
   constructor(addressOrEnsName: string) {
-    super(`${addressOrEnsName} is not valid`, 'InvalidAddressOrEnsName');
+    super(`${addressOrEnsName} is not a valid ens name`, 'InvalidAddressOrEnsName');
```
JustynaBroniszewska: What with an address?

rzadp: Right. Changed back.

PR #1546, opened 2019-11-08
Refactor onboarding

Marik-D:


universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -36,37 +39,47 @@ export const Onboarding = (props: OnboardingProps) => {
...
-    const wallet = await walletService.deployFutureWallet(ensName, gasParameters.gasPrice.toString(), gasParameters.gasToken, showWalletCreationModal);
+    walletCreationService.onBalancePresent(() => modalService.hideModal());
+    const wallet = await walletCreationService.initiateCreationFlow(ensName);
     modalService.hideModal();
     props.onCreate && props.onCreate(wallet);
   };
 
+  const walletState = useProperty(walletService.stateProperty);
+  const transactionHash = useProperty(walletCreationService.deploymentTransactionHash);
+  const renderModal = () => {
```
JustynaBroniszewska: Extract to a separate file and make it component for example `DeploymentModal`

cehali: 👍

universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -21,6 +22,8 @@ export interface OnboardingProps {
 export const Onboarding = (props: OnboardingProps) => {
   const modalService = useModalService<ReactModalType, ReactModalProps>();
   const [walletService] = useState<WalletService>(props.walletService || new WalletService(props.sdk));
+  const [walletCreationService] = useState(() => new WalletCreationService(walletService));
```
marekkirejczyk: Should be injected (services should ALWAYS be injected!).

universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -21,6 +22,8 @@ export interface OnboardingProps {
 export const Onboarding = (props: OnboardingProps) => {
   const modalService = useModalService<ReactModalType, ReactModalProps>();
   const [walletService] = useState<WalletService>(props.walletService || new WalletService(props.sdk));
```
marekkirejczyk: Why state? Why not planning to change service during the lifecycle of the component?
Simply:
```ts
const walletService = props.walletService || new WalletService(props.sdk);
```

universal-login-react/src/ui/Onboarding/Onboarding.tsx
marekkirejczyk: 👍

universal-login-react/src/ui/Onboarding/Onboarding.tsx
Marik-D: That would create a new instance on each re-render, which is most likely not what we want

marekkirejczyk: Good point, let's make it required props then.

PR #1548, opened 2019-11-08
Backup codes prompt

rzadp:


rzadp:
Screenshot 2019-11-08 at 11 51 42
Screenshot 2019-11-08 at 11 53 33
Screenshot 2019-11-08 at 11 53 50

xitronix:
If we click 'Ok' on the promt, leave backups, backups flow will apeear anyway and show backups codes gereated screen without backups

cehali:
@xitronix The same behaviour was before and there is a task for this in trello.

xitronix:
@cehali Agree. Checked it too.

PR #1555, opened 2019-11-12
Introduce amount validation

JustynaBroniszewska:
Invalid amount for amount: 
- less or equal 0
- equal to or greater than the balance

Screenshot 2019-11-12 at 16 01 27

### DIFF
--- a/universal-login-react/src/ui/Transfer/Amount/TransferAmount.tsx
+++ b/universal-login-react/src/ui/Transfer/Amount/TransferAmount.tsx
@@ -1,7 +1,7 @@
 import React, {useState} from 'react';
 import {TransferDropdown} from './TransferDropdown';
 import {DeployedWallet} from '@universal-login/sdk';
-import {TransferDetails, TokenDetailsWithBalance, getBalanceOf, TokenDetails} from '@universal-login/commons';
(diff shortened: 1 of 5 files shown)


universal-login-react/src/ui/Transfer/Amount/TransferAmount.tsx
```diff
@@ -18,19 +18,22 @@ export interface TransferAmountProps {
...
 
   useAsyncEffect(() => sdk.subscribeToBalances(contractAddress, setTokenDetailsWithBalance), []);
   const balance = getBalanceOf(tokenDetails.symbol, tokenDetailsWithBalance);
 
   const validateAndUpdateTransferDetails = (amount: string) => {
-    if (balance && amount) {
-      setIsAmountCorrect(Number(amount) <= Number(balance));
+    if (balance && isValidAmount(amount, balance)) {
+      setValidAmount(true);
+      setEnableButton(true);
```
xitronix: Should not be single state? What is difference between validAmount and enableButton?

JustynaBroniszewska: The initial states of above are different. At the beginning you want to disable button but you don't want to show 'Invalid amount' message.

PR #1563, opened 2019-11-13
Add tilda to balance in usd and improve style

JustynaBroniszewska:
Screenshot 2019-11-13 at 15 07 51

Screenshot 2019-11-13 at 15 08 07

Screenshot 2019-11-13 at 15 08 30

Screenshot 2019-11-13 at 15 08 37


universal-login-react/src/core/utils/formatCurrency.ts
```diff
@@ -9,3 +9,8 @@ export const formatCurrency = (value: string, currency = 'USD'): string => {
   const formatter = new Intl.NumberFormat('en-US', {currency: 'USD', style});
   return formatter.format(withPrecision(Number.parseFloat(value || '0'), maximumFractionDigits));
 };
+
+export const formatUsdAmount = (amount: string, currency = 'USD') => {
```
cehali: It is called formatUsdAmount but allows pass another currency. Maybe adjust formatCurrency to return formatted amount with tilde or 0

JustynaBroniszewska: What about formatFiatAmount?

PR #1576, opened 2019-11-14
Add confirmation text to correct emoji input

cehali:
# Summary
Add confirmation text to correct emoji input


cehali:
image
image

PR #1577, opened 2019-11-14
Deprecate SDK functions

JustynaBroniszewska:


universal-login-sdk/lib/api/sdk.ts
```diff
@@ -105,18 +106,22 @@ class UniversalLoginSDK {
   }
 
   async addKey(to: string, publicKey: string, privateKey: string, executionOptions: SdkExecutionOptions): Promise<Execution> {
```
Marik-D: Add `@deprecated` annotation so linters will warn about methods being deprecated.

https://jsdoc.app/tags-deprecated.html

PR #1579, opened 2019-11-14
Move related methods implementation to DeployedWallet

JustynaBroniszewska:


universal-login-sdk/lib/api/deprecate.ts
```diff
@@ -0,0 +1,3 @@
+export function deprecateSDKMethod(subject: string) {
+  console.warn(`sdk.${subject} is deprecated. Use DeployedWallet.${subject} method instead.`);
+};
```
rzadp: Seems to me like `lib/api/` is not the right place for this little internal helper method. Maybe `core/utils`?

rzadp: Also, the name suggest that it is for deprecating ANY SDK method - but it is only for deprecating SDK methods which happen to be moved to `DeployedWallet`.

How about:
```
export function deprecateSDKMethod(subject: string, movedTo: string) {
  console.warn(`sdk.${subject} is deprecated. Use${movedTo}.${subject} method instead.`);
};
```
?

JustynaBroniszewska:
Execute function proposal:
- sdk.execute:
```ts
  async execute(message: Partial<Message>, privateKey: string): Promise<Execution> {
    ensureNotNull(this.relayerConfig, Error, 'Relayer configuration not yet loaded');
    const signedMessage = this.createDeployedWallet(message.from!, privateKey).createSignedMessage(message);
    ensure(signedMessage.gasLimit! <= this.relayerConfig!.maxGasLimit, InvalidGasLimit, `${signedMessage.gasLimit} provided, when relayer's max gas limit is ${this.relayerConfig!.maxGasLimit}`);
    ensureSufficientGas(signedMessage);
    return this.executionFactory.createExecution(signedMessage);
  }
```

- deployedWallet:
```ts
  async createSignedMessage(message: Partial<Message>): Promise<SignedMessage> {
    const nonce = message.nonce || parseInt(await this.getNonce(), 10);
    const partialMessage = {gasToken: ETHER_NATIVE_TOKEN.address, ...message, nonce};
    return messageToSignedMessage(partialMessage, this.privateKey);
  }
```

universal-login-sdk/lib/api/DeployedWallet.ts
```diff
@@ -29,6 +32,13 @@ export class DeployedWallet implements ApplicationWallet {
   ) {
   }
 
+  private getContract(): Contract {
+    if (!this.contractInstance) {
+      this.contractInstance = new Contract(this.contractAddress, WalletContractInterface, this.sdk.provider);
```
Marik-D: Why not just create it in constructor?

marekkirejczyk: Long story short - it would require tons of test refactoring, which could be done in future PRs.

PR #1589, opened 2019-11-18
Deploy stored future wallet

Marik-D:
# Summary

Fixes #number

## Checklist


universal-login-react/src/core/services/WalletCreationService.ts
```diff
@@ -1,27 +1,31 @@
...
   ) {
   }
 
-  async initiateCreationFlow(ensName: string) {
-    const {waitForBalance} = await this.walletService.createFutureWallet(ensName);
-    await waitForBalance();
-    this.callbackOnBalancePresent && this.callbackOnBalancePresent();
+  async deployWhenReady(onBalancePresent?: () => void) {
+    const state = await this.walletService.stateProperty.pipe(
+      waitFor(state => state.kind === 'Future'),
```
marekkirejczyk: Idea:
`state.isType('Future')`

universal-login-web3/package.json
```diff
@@ -11,7 +11,7 @@
     "@universal-login/relayer": "0.3.56",
     "@universal-login/sdk": "0.3.56",
     "ethers": "^4.0.39",
-    "reactive-properties": "^0.1.3",
+    "reactive-properties": "^0.1.5",
```
marekkirejczyk: Let's move it to UniversalLogin github and add license

Marik-D: It's already there and has a license

universal-login-react/src/core/services/WalletCreationService.ts
Marik-D: This would require us to turn the state into a class. I think we need to further consider the benefits and implications that would give us.

universal-login-react/test/core/services/WalletCreationService.ts
```diff
@@ -0,0 +1,62 @@
...
+    walletService = new WalletService(sdk);
+    walletCreationService = new WalletCreationService(walletService);
+  });
+
+  it('initiates creation flow', async () => {
+    await walletCreationService.initiateCreationFlow('name.mylogin.eth');
+    expect(walletService.state.kind).to.eq('Future');
+  });
+
+  describe('deployWhenReady', () => {
```
JustynaBroniszewska: What happened if we call `deployWhenReady` twice? I think that we should have a test for it. (This situation is when users cancel and go back)

universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -41,12 +41,17 @@ export const Onboarding = (props: OnboardingProps) => {
...
   const onCreateClick = async (ensName: string) => {
-    walletCreationService.onBalancePresent(() => modalService.hideModal());
-    const wallet = await walletCreationService.initiateCreationFlow(ensName);
-    modalService.hideModal();
-    props.onCreate && props.onCreate(wallet);
+    await walletCreationService.initiateCreationFlow(ensName);
   };
 
+  useEffect(() => {
+    setImmediate(async () => {
```
xitronix: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

Marik-D: Webpack should transpile it to something thats supported by the standard.

PR #1590, opened 2019-11-18
Introduce waitExpect in MessageHandler tests

JustynaBroniszewska:


rzadp:
Why? How is it better?

JustynaBroniszewska:
@rzadp we didn't wait for the message to be executed. And there were situations when tests failed because of that. Now we are waiting for them.

PR #1592, opened 2019-11-18
Add Error Message when transaction hash is empty

cehali:
# Summary
Add ErrorMessage when transaction hash is empty in ConnectionNotification


cehali:
image
image

cehali:
image
image

xitronix:
I think that if wallet will have layout like react it will be prettier

PR #1593, opened 2019-11-18
Show platform instead os in devices and remove IP address

JustynaBroniszewska:
**BEFORE**
Screenshot 2019-11-18 at 16 03 59

**AFTER**
Screenshot 2019-11-18 at 16 08 14

Screenshot 2019-11-18 at 16 08 09


xitronix:
Should we change device icon if it android for example?

PR #1594, opened 2019-11-19
Rename gasLimitExecution to gasCall and gasData to gasBase

JustynaBroniszewska:


xitronix:
I would prefer remain 'Limit' in name

universal-login-commons/lib/core/utils/messages/computeGasData.ts
```diff
@@ -16,3 +16,5 @@ export const computeGasData = (data: string) => {
 const gasCostFor = (byte: string) => {
   return byte === '00' ? ZERO_BYTE_GAS_COST : NON_ZERO_BYTE_GAS_COST;
 };
+
+export const computeGasBase = computeGasData;
```
marekkirejczyk: Clever, I see what you did there :)
 Still I would do:
```ts
export const computeGasBase = (data: string) => {
  return computeGasData;
};
```
A little bit more self explanatory.

PR #1597, opened 2019-11-19
Handle gasPrice is equal 0

xitronix:


marekkirejczyk:
What about validation on the front-end?

xitronix:
It shoudn't happen, because we should block action buttons when gasPriceComponent doesn't loaded yet

PR #1606, opened 2019-11-22
Add different validation for ether and token

cehali:
# Summary
Add different validation for ether and token


universal-login-commons/lib/core/utils/isValidAmount.ts
```diff
@@ -1,14 +1,23 @@
...
-    }
+export const isValidAmount = (tokenAddress: string, balance: Nullable<string>, gasCostInWei: utils.BigNumber, amount?: string): boolean => {
+  if (balance) {
+    const maxEtherAmount = utils.parseEther(balance).sub(gasCostInWei);
+    return tokenAddress === ETHER_NATIVE_TOKEN.address
+      ? isAmountInConstraints(maxEtherAmount, amount)
+      : isAmountInConstraints(utils.parseEther(balance), amount);
+  }
+  return false;
+};
```
marekkirejczyk:
```ts
export const isValidAmount = (
   tokenAddress: string, balance: Nullable<string>, 
   gasToken: string,
   gasCostInWei: utils.BigNumber, 
   amount?: string): boolean => {

if (!balance) {
  return false;
}
const maxEtherAmount = utils.parseEther(balance).sub(gasCostInWei);
if (tokenAddress === gasToken) {
      return isAmountInConstraints(maxEtherAmount, amount)
} 
return isAmountInConstraints(utils.parseEther(balance), amount);
```

universal-login-commons/lib/core/utils/isValidAmount.ts
```diff
@@ -1,14 +1,23 @@
...
+
+export const isAmountInConstraints = (balance: utils.BigNumber, amount?: string): boolean => {
+  if (amount && amount.match(isNumber)) {
+    const amountAsBigNumber = utils.parseEther(amount);
+    return amountAsBigNumber.gt(0) && amountAsBigNumber.lt(balance);
```
marekkirejczyk: lt -> lte

universal-login-react/src/ui/Transfer/Transfer.tsx
```diff
@@ -24,27 +24,20 @@ interface ErrorsProps {
   recipientError: boolean;
 }
 
-export const Transfer = ({deployedWallet, transferDetails, updateTransferDetailsWith, tokenDetails, onSendClick, transferClassName}: TransferProps) => {
+export const Transfer = ({deployedWallet, transferDetails, updateTransferDetailsWith, tokenDetailsWithBalance, tokenDetails, onSendClick, transferClassName}: TransferProps) => {
```
marekkirejczyk: We can now simplify and simply pass the transferService to Transfer component.
This way we can get rid of `deployedWallet`, `tokenDetails`, `tokenDetailsWithBalance `, `onSendClick` (inline), `transferDetails` (move from props to state), updateTransferDetailsWith (move to current component).

universal-login-react/src/ui/Transfer/Transfer.tsx
```diff
@@ -24,27 +24,20 @@ interface ErrorsProps {
...
-    areInputsValid(errorsCalculated) && onSendClick();
+  const onTransferClick = async () => {
+    try {
+      await onSendClick();
+    } catch (error) {
+      setErrors({
+        amountError: error.errorType !== 'InvalidAddressOrEnsName' && true,
+        recipientError: error.errorType !== 'InvalidAmount' && true,
+      });
+    }
```
marekkirejczyk: We can now move ErrorProps to app layer and use them as a state instead as props.
That way we can have:
```ts
const {errors} = await transferService.transfer(transferDetails);
if (errors) {
  setErrors(errors);
} else {
  modalService.showModal('transfer');
}
```

or even better:
```ts
  history.push('/transfer')
```

xitronix: I planned change whole transfer flow into /transfer route and flow like in BackupCodes after this pr

PR #1608, opened 2019-11-25
Introduce calculateGasBase

JustynaBroniszewska:


universal-login-contracts/lib/message.ts
```diff
@@ -17,11 +17,12 @@ export const messageToUnsignedMessage = (message: Partial<Message>): UnsignedMes
     nonce: message.nonce!,
     gasPrice: message.gasPrice!,
     gasToken: message.gasToken!,
-    gasBase: 0,
-    gasCall: 0,
+    gasLimit: message.gasLimit!,
```
xitronix: As I understand we can update messageToUnsignedMessage signature to `(message:Message): UnsignedMessage`,
because this message used only in `calculateGasBase` where type of arg is `Omit<Message, 'gasLimit'>` and all fields in variable `message` are required. So we also don't need gasLimit in messageWithoutGasEstimates

JustynaBroniszewska: We need `gasLimit` to calculate `gasCall`. See line 23.

JustynaBroniszewska:
```suggestion
```

universal-login-contracts/lib/message.ts
```diff
@@ -17,11 +17,10 @@ export const messageToUnsignedMessage = (message: Partial<Message>): UnsignedMes
     nonce: message.nonce!,
     gasPrice: message.gasPrice!,
     gasToken: message.gasToken!,
```
xitronix: Should we check if these fields exist or update function's signature?

xitronix: With such types we will have consistent types in this functions:
```
export const messageToUnsignedMessage = (message: Message & Partial<Pick<Message, 'value' |  'data'>>): UnsignedMessage => {
  const messageWithoutGasEstimates =  {
    ...message,
    value: message.value || utils.bigNumberify(0),
    data: message.data || '0x',
    gasBase: 0,
    gasCall: 0,
  }
```

PR #1612, opened 2019-11-26
Integrate Wyre in topUp

cehali:
# Summary
Integrate Wyre v2


universal-login-react/src/ui/TopUp/Fiat/FiatFooter.tsx
```diff
@@ -53,7 +53,6 @@ export const FiatFooter = ({paymentMethod}: FiatFooterProps) => {
     case TopUpProvider.WYRE:
       return (
         <div className="info-block info-row">
-          <p className="info-text info-text-hint">Available soon</p>
         </div>
```
JustynaBroniszewska: #nextPR 

we can add icons here, for example, Apple pay icon

universal-login-react/src/integration/Wyre.tsx
```diff
@@ -0,0 +1,21 @@
+import {WyreConfig} from '@universal-login/commons';
+
+interface WyreProps {
+  address: string;
+  currency: string;
+  config: WyreConfig;
+}
+
+export const Wyre = ({address, currency, config}: WyreProps) => {
```
marekkirejczyk: Wyre.tsx along with Ramp nad Saffello

universal-login-react/src/ui/TopUp/Fiat/FiatFooter.tsx
marekkirejczyk: 👍

PR #1627, opened 2019-12-03
Refactor WalletService to support refresh function

xitronix:


universal-login-sdk/lib/core/utils/errors.ts
```diff
@@ -215,7 +215,7 @@ export class Overridden extends SDKError {
 
 export class WalletOverridden extends Overridden {
   constructor() {
-    super('Wallet cannot be overridded', 'WalletOverridden');
+    super('Wallet cannot be overridden', 'WalletOverridden');
```
JustynaBroniszewska: 👍

PR #1635, opened 2019-12-03
Update Wallet States uml

JustynaBroniszewska:


docs/modeling/src/concepts/walletServiceStates.pml
```diff
@@ -1,23 +1,24 @@
...
-None --> Future
-None --> Connecting
-Connecting --> Deployed
-Future --> Deploying
-Deploying --> Future
-Deploying --> Deployed
+None --> Future: createFuture
+None --> Connecting: connect
+Connecting --> Deployed: waitForSuccess
+Future --> Deploying: waitForDeploy
```
xitronix: Future --> None: disconnect

JustynaBroniszewska: We can disconnect from each state... Adding this can make this image a little unreadable.

PR #1638, opened 2019-12-04
Replace WalletCreationService in favor of WalletService

xitronix:


universal-login-react/src/App.tsx
```diff
@@ -58,16 +58,31 @@ export const App = () => {
...
-              <Onboarding
-                sdk={sdk}
-                walletService={walletService}
-                domains={['mylogin.eth', 'universal-id.eth']}
-                tryEnablingMetamask={tryEnablingMetamask}
-                onConnect={() => console.log('connected')}
-                onCreate={() => console.log('created')}
-              />
-            </Route>
+            <Route
```
cehali: For me, this Route is not working

universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -52,20 +36,28 @@ export const Onboarding = (props: OnboardingProps) => {
                   <div className="perspective">
                     <WalletSelector
                       sdk={props.sdk}
-                      onCreateClick={onCreateClick}
+                      onCreateClick={async (ensName) => {
```
cehali: Minor thing, I think when function passed to a component is longer than one line it should be extracted

universal-login-react/src/ui/Onboarding/Onboarding.tsx
```diff
@@ -52,20 +36,28 @@ export const Onboarding = (props: OnboardingProps) => {
...
+                        await props.walletService.createFutureWallet(ensName);
+                        history.push('/create');
+                      }}
```
cehali: Minor thing, I think when function passed to a component is longer than one line it should be extracted

xitronix: I did it in such way just because i wanted to have history prop from RouteProps and thought that it is simplier and more readable than create function wit history arg.

PR #1656, opened 2019-12-06
ConnectAccount use Routes

JustynaBroniszewska:


universal-login-wallet/src/app/constants/urlMapping.ts
```diff
@@ -3,6 +3,8 @@ export const urlMapping: Record<string, string[]> = {
     '/welcome',
     '/privacy',
     '/terms',
+    '/connect',
```
xitronix: Is it necessary?

PR #1662, opened 2019-12-06
Move example web3

xitronix:


xitronix:
@JustynaBroniszewska this move will done in another way. Because react can't have  dependencies on @universal-login/web3. So i will move example in @universal-login/web3 or just remove it.
Another option can be moving whole playground to example but i consider we should try move it to web3

PR #1665, opened 2019-12-09
Show minimal ETH amount in TopUpWithFiat

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/Fiat/FiatFooter.tsx
```diff
@@ -29,6 +30,9 @@ export const FiatFooter = ({paymentMethod}: FiatFooterProps) => {
           <div className="info-block info-row">
             <p className="info-text info-text-hint">Minimum amount is 1£</p>
           </div>
+          <div className="info-block info-row">
+            <p className="info-text info-text-hint">Minimum amount required to deploy contract is {minimalAmount} ETH</p>
+          </div>
```
xitronix: We should do not show this info if it is not deployment and probably use same info as in crypto (DeploymentWithCryptoInfo)

cehali: 👍

universal-login-react/src/ui/TopUp/TopUpWithCrypto.tsx
```diff
@@ -55,7 +50,7 @@ export const TopUpWithCrypto = ({contractAddress, isDeployment, minimalAmount}:
             />
           </div>
           {isDeployment
-            ? <DeploymentWithCryptoInfo minimalAmount = {minimalAmount}/>
+            ? <MinimalAmountInfo minimalAmount={minimalAmount} paymentMethod='crypto' />
             : <TopUpCryptoInfo/>
```
JustynaBroniszewska: This one should be also shown in the deployment?

cehali: It hasn't been till today so I don't think so

PR #1681, opened 2019-12-17
Extend transfer validators

cehali:
# Summary


universal-login-sdk/lib/core/services/validations/RecipientValidator.ts
```diff
@@ -1,28 +1,37 @@
 import {Validator, TransferErrors} from './Validator';
 import {utils} from 'ethers';
 import {isValidEnsName, TransferDetails} from '@universal-login/commons';
+import UniversalLoginSDK from '../../../api/sdk';
+import {AddressZero} from 'ethers/constants';
 
-function isProperAddress(recipient: string): boolean {
+const isProperAddress = (recipient: string): boolean => {
```
xitronix: Merge with isProperAddress in commons

universal-login-wallet/test/integration/ui/Transfer.tsx
```diff
@@ -50,7 +50,7 @@ describe('UI: Transfer', () => {
       : '0x0';
   });
 
-  it('send ETH => invalid ensName', async () => {
+  xit('send ETH => invalid ensName', async () => {
```
xitronix: remove xit

universal-login-sdk/lib/core/services/validations/RecipientValidator.ts
cehali: isProperAddress in commons looks like this:
```ts
export const isProperAddress = (address: string) => {
  return !!address.match(/^0x[0-9A-Fa-f]{40}$/);
};
```
And here we use ethers to define if address is valid.

universal-login-wallet/test/integration/ui/Transfer.tsx
cehali: 👍

universal-login-sdk/lib/core/services/validations/RecipientValidator.ts
cehali: I can update isProperAddress in commons in next PR

PR #1700, opened 2019-12-19
Move */lib/* => */src/*

xitronix:


universal-login-relayer/src/http/relayers/DevelopmentRelayer.ts
```diff
@@ -1,7 +1,7 @@
 import {utils, Contract, providers} from 'ethers';
 import {waitToBeMined} from '@universal-login/commons';
 import {Config} from '../../config/relayer';
-import Token from './abi/Token.json';
+import Token from './abi/Token.json.js';
```
JustynaBroniszewska: Is it correct?

universal-login-relayer/test/helpers/startRelayer.js
```diff
@@ -1,11 +1,11 @@
 import {deployContract} from 'ethereum-waffle';
 import WalletContract from '@universal-login/contracts/build/Wallet.json';
 import {deployFactory} from '@universal-login/contracts';
-import Token from '../../lib/http/relayers/abi/Token.json';
+import Token from '../../src/http/relayers/abi/Token.json.js';
```
JustynaBroniszewska: same

PR #1704, opened 2019-12-19
Add minimalAmount to relayer

cehali:
# Summary


universal-login-relayer/src/integration/ethereum/WalletService.ts
```diff
@@ -31,6 +34,11 @@ class WalletService {
     this.hooks.emit('created', {transaction, contractAddress});
     return transaction;
   }
+
+  getTokensWithMinimalAmount(gasPrice: string) {
+    return this.supportedTokens.map((supportedToken) =>
+      ({...supportedToken, minimalAmount: safeMultiply(MINIMAL_DEPLOYMENT_GAS_LIMIT, gasPrice)}));
```
xitronix: Every token has gasPrice in own currency

cehali: Yes, but currently we assume that for each token minimal amount is the same.

PR #1709, opened 2019-12-20
Adjust migration path in startDevelopment

cehali:
# Summary
Adjust migration path in startDevelopment


universal-login-ops/src/dev/startDevelopment.ts
```diff
@@ -96,7 +96,7 @@ function getProxyContractHash() {
 
 function getMigrationPath() {
   const packagePath = require.resolve('@universal-login/relayer/package.json');
-  return join(dirname(packagePath), 'build', 'lib', 'integration', 'sql', 'migrations');
+  return join(dirname(packagePath), 'dist/cjs', 'src', 'integration', 'sql', 'migrations');
```
xitronix: be constistent and use 'dist', 'cjs'?

PR #1723, opened 2019-12-26
[1] Introduce reactive observers

marekkirejczyk:


universal-login-sdk/src/core/properties/ReferenceCountedProperty.ts
```diff
@@ -0,0 +1,25 @@
+import {State} from 'reactive-properties';
+import {Callback} from 'reactive-properties/dist/Property';
```
JustynaBroniszewska: We should change that as soon as [this PR](https://github.com/UniversalLogin/reactive-properties/pull/1) is merged.

PR #1724, opened 2019-12-27
Use mocked contracts from @contracts/testutils

xitronix:


universal-login-contracts/src/ierc20/index.ts
```diff
@@ -0,0 +1,3 @@
+import IERC20 from './IERC20.json';
```
marekkirejczyk: I propose directory name: `/src/token/`

PR #1777, opened 2020-01-16
Generate just one backup code

JustynaBroniszewska:
I didn't change the interface of backup codes services (BackupCodesService or BackupCodesWithExecution). `codes` is still an array of codes, but now it has only one code. Additionally, `generateBackupCodes` function uses `addKey` function instead of using `addKeys` function that will not work with GnosisSafe.

### DIFF
--- a/universal-login-react/test/ui/BackupCodes.test.tsx
+++ b/universal-login-react/test/ui/BackupCodes.test.tsx
@@ -33,7 +33,7 @@ describe('INT: BackupCodes', () => {
     });
     dashboard.backupCodes().clickGenerate();
     await waitExpect(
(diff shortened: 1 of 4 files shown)


universal-login-react/test/ui/BackupCodes.test.tsx
```diff
@@ -33,7 +33,7 @@ describe('INT: BackupCodes', () => {
     });
     dashboard.backupCodes().clickGenerate();
     await waitExpect(
-      () => expect(dashboard.backupCodes().getBackupCodes().length).to.be.eq(2),
+      () => expect(dashboard.backupCodes().getBackupCodes().length).to.be.eq(1),
       13000,
     );
   }).timeout(15000);
```
xitronix: I think we can remove timeouts now

PR #1789, opened 2020-01-21
SDK: Rename Properties => States

xitronix:
We should rename Properties => State, because they have interface which extended from State and give us State's functionality.


universal-login-sdk/src/core/states/BlockState.ts
```diff
@@ -0,0 +1,18 @@
+import {BlockchainService} from '@universal-login/contracts';
+import {ReferenceCountedState} from './ReferenceCountedState';
+
+export class BlockState extends ReferenceCountedState<number> {
```
Marik-D: `BlockNumberState` would probably be more descriptive

xitronix: I think it will be BlockNumberService in the end of the day, so i will rename it later.

PR #1791, opened 2020-01-22
[ Contracts, Sdk ]: Extend blockchain service

xitronix:


universal-login-contracts/src/integration/BlockchainService.ts
```diff
@@ -20,6 +20,14 @@ export class BlockchainService {
     return this.provider.getLogs(filter);
   }
 
+  on(eventType: providers.EventType, listener: providers.Listener) {
```
JustynaBroniszewska: Consider another name e.g. `onEvent`. Name `on` doesn't say a lot.

PR #1793, opened 2020-01-22
Make DisconnectAccountService test wait for 'None' WalletState

xitronix:


universal-login-react/test/core/services/DisconnectAccountService.test.ts
```diff
@@ -24,6 +25,7 @@ describe('DisconnectAccountService', () => {
   before(async () => {
     ([wallet] = getWallets(createMockProvider()));
     ({deployedWallet, relayer} = await setupDeployedWallet(wallet, ensName));
+    (deployedWallet.sdk.provider as any).pollingInterval = 100;
```
JustynaBroniszewska: why not 5?

xitronix: This test will fail with 5 or 10 with timeout

PR #1800, opened 2020-01-23
Organize  Storage Services

xitronix:


universal-login-sdk/src/core/models/WalletService.ts
```diff
@@ -41,6 +41,12 @@ export type SerializedWalletState = {
   wallet: ApplicationWallet;
 };
 
+export interface IStorageService {
```
JustynaBroniszewska: Move it to `models/StorageService`

xitronix: 👍

PR #1802, opened 2020-01-24
Extract block storage

xitronix:


universal-login-sdk/src/core/observers/WalletEventsObserverFactory.ts
```diff
@@ -39,6 +61,7 @@ class WalletEventsObserverFactory {
   stop() {
     ensureNotNull(this.unsubscribe, InvalidObserverState);
     this.unsubscribe();
+    this.unsubscribe = null;
```
JustynaBroniszewska: why?

xitronix: We have not need unsubscribe  already, so I just nullify it and ensure at the 39th line now make sens

PR #1827, opened 2020-01-31
Web3: Show recipient and value on Transaction confirmation

xitronix:


universal-login-web3/src/ui/react/Confirmation/TransactionSpeed.tsx
```diff
@@ -1,29 +1,33 @@
...
+export const TransactionSpeed = ({gasModes, selectedValue, onChange}: TransactionSpeedProps) => {
+  return (
+    <Row>
+      {gasModes.map(({name, usdAmount}) => (
+        <RadioButton
+          key={name}
+          value={name}
+          selectedValue={selectedValue}
+          onChange={onChange}
+          time="1 day"
```
JustynaBroniszewska: Is 1 day correct for each gasMode?

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -165,7 +165,7 @@ export class ULWeb3Provider implements Provider {
     this.uiController.finishOnboarding();
   }
 
-  private async ensureWalletIsDeployed() {
+  private async deployWalletIfNotExists() {
```
JustynaBroniszewska: Rename proposals:
- `deployIfNoWallet`
- `deployIfNoWalletDeployed`
- `deployIfNoWalletExist`

xitronix: Done

universal-login-web3/src/ui/react/Confirmation/TransactionSpeed.tsx
xitronix: It is mock up, so i just write value. We don't have any service which calculate it. Which value do you suggest write here?

universal-login-web3/src/ui/react/Confirmation/TransactionConfirmation.tsx
```diff
@@ -0,0 +1,151 @@
...
+                <DataLabel>Send to:</DataLabel>
+                <Address>{transaction.to}</Address>
+              </Row>
+              <Row>
+                <DataLabel>Value:</DataLabel>
+                <ValueRow>
+                  <Highlighted>
+                    <Value>{utils.formatEther(transaction.value)} ETH</Value>
+                  </Highlighted>
+                  <Value>{mode.usdAmount}</Value>
```
marekkirejczyk: If not working better hide

PR #1839, opened 2020-02-05
Simplify initiating UL button in Web3

JustynaBroniszewska:


universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -175,8 +176,9 @@ export class ULWeb3Provider implements Provider {
     }
   }
 
-  initWeb3Button(element: Element) {
-    renderLogoButton(element, this.walletService);
+  initWeb3Button() {
+    const element = getOrCreateUlButton();
```
xitronix: Change to get or add element injection

PR #1850, opened 2020-02-06
Move isDataForFunctionCall to commos

xitronix:


universal-login-commons/src/core/utils/isDataForFunctionCall.ts
```diff
@@ -0,0 +1,6 @@
+import {utils} from 'ethers';
+
+export const isDataForFunctionCall = (data: string, contract: any, functionName: string) => {
```
Marik-D:
```suggestion
export const isDataForFunctionCall = (data: string, contract: Contract, functionName: string) => {
```

Marik-D: pass `utils.Interface` instead of contract

universal-login-commons/src/core/utils/isDataForFunctionCall.ts
```diff
@@ -0,0 +1,6 @@
...
+  const functionSignature = new utils.Interface(contract.interface).functions[functionName].sighash;
+  return functionSignature === data.slice(0, functionSignature.length);
```
Marik-D:
```suggestion
  return data.startsWith(functionSignature);
 
```

PR #1865, opened 2020-02-07
Add id for deployments

xitronix:


xitronix:
This pr remain all logic. When we use  id instead of hash we will remove this rows' length check

JustynaBroniszewska:
Close, for now, will continue in the near future

PR #1867, opened 2020-02-07
Update messages migration (after #1865)

xitronix:
after #1865


JustynaBroniszewska:
Close, for now, will continue in the near future

PR #1902, opened 2020-02-17
Add notifications indicator to uniButton

xitronix:


universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -99,9 +112,17 @@ export class ULIFrameProvider {
     this.setDashboardVisibility(false);
   }
 
+  private setElementVisibility = (element: HTMLDivElement, isVisible: boolean) => {
```
Marik-D: Can be reduced to something like https://github.com/UniLogin/UniLogin/blob/master/universal-login-provider/src/ULIFrameProvider.ts#L74

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -57,11 +58,23 @@ export class ULIFrameProvider {
       case 'ul_ready':
         this.isReady.set(true);
         break;
+      case 'ul_set_notification_indicator':
+        this.isNewNotifications.set(msg.params[0]);
+        this.setNotificationsIndicator(this.isNewNotifications.get());
```
Marik-D: Instead of calling this function every time the state is set, you can just subscribe to the state in `initUlButton`.
I've added a possible implementation below

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -114,7 +135,20 @@ export class ULIFrameProvider {
 
     element.innerHTML = `
       <img src="${this.config.logoUrl}" alt="U" >
+      <div id='uni-notifications'></div>
     `;
+    const notificationIndicator = element.getElementsByTagName('div')[0];
```
Marik-D: Adding something like this here should simplify the code above
```typescript
this.isNewNotifications.pipe(forEach(hasNotifications => { notificationIndicator.style.display = hasNotifications ? 'block' : 'none' }))
```
This would also make `setNotificationsIndicator` method unneeded

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -13,6 +13,7 @@ export class ULIFrameProvider {
   private iframe: HTMLIFrameElement;
   private bridge: RpcBridge;
   private isReady = new State(false);
+  private isNewNotifications = new State(false);
```
Marik-D: Rename to `hasNotifications`

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -99,9 +112,17 @@ export class ULIFrameProvider {
...
+      Object.assign(element.style, {display: 'block'});
+    } else {
+      Object.assign(element.style, {display: 'none'});
+    }
+  };
+
   private boundOpenDashboard = this.openDashboard.bind(this);
 
-  initUlButton(element: HTMLButtonElement) {
+  initUlButton(element: HTMLButtonElement, isNewNotifications: boolean) {
```
Marik-D: Since you have access to `this.isNewNotifications` you can omit this parameter and just access the field directly

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -40,9 +40,11 @@ export class ULWeb3Provider implements Provider {
   private readonly sdk: UniversalLoginSDK;
   private readonly walletService: WalletService;
   private readonly uiController: UIController;
+  private unsubscribeNotifications: Nullable<Function> = null;
 
   readonly isLoggedIn: Property<boolean>;
   readonly isUiVisible: Property<boolean>;
+  readonly isNewNotifications: State<boolean> = new State(false);
```
Marik-D: I think the code would be better if this property would be a part of `DeployedWallet`'s interface. So something like:
```typescript
class DeployedWallet {
...
  private readonly pendingAuthorizationsState = new State<Notification[]>()
  public  readonly pendingAuthorizations = pendingAuthorizationsState.pipe(
    withEffect(() => this.subscribeAuthorizations(x => this.pendingAuthorizationsState.set(x))),
  )
...
}
```

Than `isNewNotifications` can be declaratively expressed as:
```typescript
readonly isNewNotifications = this.walletService.walletState.pipe(
  map(walletState => {
    if(walletState.kind === 'Deployed') {
       return walletState.deployedWallet.pendingAuthorizations.pipe(
             map(pendingAuthorizations => pendingAuthorizations.length > 0)
       )
     } else {
       return new State(false) // we cannot just return `false` here since the other brach of the `if` statement has the type `Property<boolean>`
    }
  }),
  flat,  // the `map` function above returned `Property<Property<boolean>>` which we can reduce to `Property<boolean>` using the `flat` operator
)
```

xitronix: i think we don't want by default subscribe AuthorisationObserver. If we would it will be next PR

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -57,11 +58,23 @@ export class ULIFrameProvider {
...
+        this.hasNotifications.set(msg.params[0]);
+        this.setNotificationsIndicator(this.hasNotifications.get());
+        break;
       default:
         this.sendUpstream(msg, cb);
     }
   }
 
+  private setNotificationsIndicator(hasNotifications: boolean) {
+    const uniButton = document.getElementById('unilogin-button');
```
JustynaBroniszewska: `unilogin-button` should be extracted

universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -114,7 +131,20 @@ export class ULIFrameProvider {
 
     element.innerHTML = `
       <img src="${this.config.logoUrl}" alt="U" >
+      <div></div>
```
JustynaBroniszewska: Why not use id e.g. `unilogin-button-notification`?

xitronix: I removed id. Because when i want operate on unilogin-button-notification only when button exists. So i get this element from ulButtonElement

universal-login-web3/src/ULWeb3Provider.ts
Marik-D: The code above will only subscribe to authorisations when somebody starts observing the property. I think it would be better to do it in this PR so not to introduce the complex subscription handling logic in `ULWeb3Provider` which is already has a lot of stuff in it.

PR #1904, opened 2020-02-18
Speed up message handling (reduce ~8sec)

xitronix:
Additional information in trello's ticket

### DIFF
--- a/universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
+++ b/universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
@@ -26,7 +26,6 @@ export default class PendingMessages {
     }
     await this.addSignatureToPendingMessage(messageHash, message);
     const status = await this.getStatus(messageHash);
-    status.messageHash = messageHash;
     if (this.isEnoughSignatures(status)) {
       await this.onReadyToExecute(messageHash, status);
     }
--- a/universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
+++ b/universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
@@ -27,14 +27,14 @@ export default class PendingMessages {
     await this.addSignatureToPendingMessage(messageHash, message);
     const status = await this.getStatus(messageHash);
     status.messageHash = messageHash;
-    if (await this.isEnoughSignatures(messageHash)) {
-      await this.onReadyToExecute(messageHash, message);
+    if (await this.isEnoughSignatures(status)) {
+      await this.onReadyToExecute(messageHash, status);
     }
     return status;
   }
 
-  private async onReadyToExecute(messageHash: string, message: SignedMessage) {
-    await this.ensureCorrectExecution(messageHash);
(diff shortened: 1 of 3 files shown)


universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
```diff
@@ -53,14 +53,14 @@ export default class PendingMessages {
...
+  async ensureCorrectExecution(messageStatus: MessageStatus) {
+    const {required, transactionHash, totalCollected} = messageStatus;
     ensure(!transactionHash, DuplicatedExecution);
-    ensure(await this.isEnoughSignatures(messageHash), NotEnoughSignatures, required, totalCollected);
+    ensure(await this.isEnoughSignatures(messageStatus), NotEnoughSignatures, required, totalCollected);
   }
 
-  async isEnoughSignatures(messageHash: string): Promise<boolean> {
-    const {totalCollected, required} = await this.getStatus(messageHash);
+  async isEnoughSignatures(messageStatus: MessageStatus): Promise<boolean> {
```
Marik-D: No need for `async`

universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
```diff
@@ -53,14 +53,14 @@ export default class PendingMessages {
     return this.statusService.getStatus(messageHash);
   }
 
-  async ensureCorrectExecution(messageHash: string) {
-    const {required, transactionHash, totalCollected} = await this.statusService.getStatus(messageHash);
+  async ensureCorrectExecution(messageStatus: MessageStatus) {
```
Marik-D: No need for `async` (since `isEnoughSignatures` can also be sync)

universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
```diff
@@ -27,14 +27,14 @@ export default class PendingMessages {
     await this.addSignatureToPendingMessage(messageHash, message);
     const status = await this.getStatus(messageHash);
     status.messageHash = messageHash;
```
JustynaBroniszewska: Unnecessary line

rzadp: Agree, looks unnecessary.
But if it is in fact necessary, then I feel that `getStatus` should be modified to include `messageHash` in the status

rzadp:
Looks good to me as well.
Not sure about the removal of `await` before `this.server.close();` though.

universal-login-relayer/src/core/services/execution/messages/PendingMessages.ts
rzadp: Also note that this was not introduced in this PR.

JustynaBroniszewska: But still, we can remove it

JustynaBroniszewska:
```suggestion
```

PR #1906, opened 2020-02-19
Detect incognitoMode

xitronix:


universal-login-wallet/src/index.tsx
```diff
@@ -1,38 +1,41 @@
...
+  if (!params) {
     return (
       <div id="preloader">
         <img src={Logo} className="preloaderImg" />
       </div>
     );
   }
+  const {services, isIncognito} = params;
+
+  isIncognito && alert('Warning! Please do not use incognito mode.');
```
JustynaBroniszewska: Maybe we should add some information about why the user shouldn't use incognito. `Warning! Please do not use incognito mode. You can lose all your funds.`

PR #1924, opened 2020-02-24
Replace ensureNotNull with ensureNotNullish

xitronix:
Should fix missing parameter error in web3 integration also


xitronix:
image

when value is undefined

PR #1931, opened 2020-02-25
Update ErrorMessage styles

JustCaptcha:


JustCaptcha:
image
Don't look to the background. When ErrorMessaged wrapped with ModalWrapper, it'll be ok.

universal-login-react/src/App.tsx
```diff
@@ -221,6 +222,10 @@ export const App = () => {
                   }
                 }}>
               </Route>
+              <Route
+                exact path="/errorMessage"
+                render={() => <ErrorMessage/>}
+              ></Route>
```
JustynaBroniszewska: It can be:
```ts
<Route
                exact path="/errorMessage"
                render={() => <ErrorMessage/>}
              />
```

JustCaptcha:
Legacy
image

UniLogin
image

Jarvis
image

PR #1947, opened 2020-02-28
Display only up to 9 decimal places of cryptocurrency

JustCaptcha:


universal-login-react/src/core/utils/formatCrypto.ts
```diff
@@ -0,0 +1,3 @@
+export const formatCrypto = (crypto: string, places: number) => {
+  return parseFloat(crypto).toFixed(places).toString();
```
xitronix: crypto: utils.BigNumberish
return utils.formatUnits(utils.bigNumberify(value.toString()), number)

JustynaBroniszewska: Add test for it!!!!

universal-login-react/src/core/utils/formatCryptoBalance.ts
```diff
@@ -0,0 +1,6 @@
+import {utils} from 'ethers';
+
+export const formatCryptoBalance = (crypto: string, places: number) => {
+  const base = utils.bigNumberify(10).pow(places);
+  return utils.formatEther(utils.bigNumberify(utils.parseEther(crypto)).div(base).mul(base));
```
xitronix:
```suggestion
  return utils.formatEther(utils.parseEther(crypto).div(base).mul(base));
```

PR #1964, opened 2020-03-04
Ignore random Infura message

JustynaBroniszewska:


universal-login-web3/src/services/UIController.ts
```diff
@@ -65,6 +65,9 @@ export class UIController {
   }
 
   showError(errorMessage?: string) {
```
xitronix: Should we ignore it? Maybe we should resend requsest in this case or at least console this error?

PR #1970, opened 2020-03-04
Introduce normalizeUpstream in ULIFrameProvider

JustynaBroniszewska:


universal-login-provider/src/ULIFrameProvider.ts
```diff
@@ -103,28 +104,17 @@ export class ULIFrameProvider {
   static create(network: Network, config = DEFAULT_CONFIG) {
     return new ULIFrameProvider({
       enablePicker: false,
-      network: network.toString() as Network,
+      ...normalizeUpstream(network),
```
Marik-D:
```suggestion
            network: network.toString() as Network,	
```

PR #1988, opened 2020-03-06
Make web3 package public

JustynaBroniszewska:


JustynaBroniszewska:
Let's leave it private for now

PR #1991, opened 2020-03-09
Enable connecting to unitest.eth

JustynaBroniszewska:


universal-login-react/src/app/filterENSDomains.ts
```diff
@@ -0,0 +1,11 @@
+import {parseDomain} from '@unilogin/commons';
+
+const DEPRECATED_DOMAINS = ['unitest.eth'];
```
xitronix: Suggestion: CREATION_NOT_SUPPORTED_DOMAINS, CREATE_DEPRECATED_DOMAINS??

universal-login-react/src/app/filterENSDomains.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const filterENSDomains = (domains: string[]) => domains.filter((domain) => !DEPRECATED_DOMAINS.includes(domain));
```
xitronix: We don't use it

PR #2005, opened 2020-03-12
Update ULWeb3Provider initialization

xitronix:


universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -90,27 +90,23 @@ export class ULWeb3Provider implements Provider {
       walletService: this.walletService,
       uiController: this.uiController,
     });
-
-    if (this.browserChecker.isLocalStorageBlocked()) {
-      this.uiController.showLocalStorageWarning();
-    } else {
-      this.walletService.loadFromStorage();
-    }
```
JustynaBroniszewska: what will happen if we already have deployed wallet?

Marik-D:
### Feedback regarding general ULWeb3Provider initialization logic:

1. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR95
This has a race condition. Since is setting `initialized = true` is only done at the end nothing prevents SDK from being initialized twice. I suggest using `Lazy<T>` to handle this.

2. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR233 Another similar race condition here 

3. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR108 If we skip initialization because of blocked local storage, this wont stop the flow, even tho it should. We should handle this somehow 

4. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR103
This is duplicated here and in `UIController`. I suggest just creating a single property in `ULWeb3Provider` and passing it in to `UIController`:

5. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR109
This is probably not needed since wallet initialization is already handled in specific RPC handler methods. Likewise, `methodsRequiringAccount` wont be required also. 

6.  https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fL152
I suggest renaming this to `ul_stop` to prevent confusion with connecting/disconnection to UniLogin wallet with another key.

7. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR228
This method is not used and should be removed.

8. https://github.com/UniLogin/UniLogin/pull/2005/files#diff-3057f6bebc8dde3f4fb1348ec84c519fR133
Logic here should be different: `eth_accounts` shouldn't trigger the onboarding if the user doesn't have a wallet yet and just return `[]`. `eth_requestAccounts` should trigger the onboarding, wait, and then return an array with users new contract address.

universal-login-commons/test/core/utils/InitializationHandler.unit.test.ts
```diff
@@ -0,0 +1,110 @@
+import chai, {expect} from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import {InitializationHandler} from '../../../src/core/utils/InitializationHandler';
+
+chai.use(sinonChai);
+
+describe('UNIT: InitializationHandler', () => {
+  describe('sync', () => {
+    const initialize = sinon.stub();
```
Marik-D: Make those into a promises that resolve after some time (like 10 ms) to test for race conditions

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -47,7 +47,7 @@ export class ULWeb3Provider implements Provider {
 
   readonly isLoggedIn: Property<boolean>;
   readonly isUiVisible: Property<boolean>;
-  readonly hasNotifications: Property<boolean>;
+  readonly hasNotifications: Property<boolean>;;
```
Marik-D:
```suggestion
  readonly hasNotifications: Property<boolean>;
```

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export class InitializationHandler<T, B> {
+  private state: State<T, B> = {
+    kind: 'Finalized',
+    value: undefined,
+  };
+
+  constructor(
+    private _initialize: () => T,
```
Marik-D: Do we use the pattern with underscores anywhere else in the project?

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -90,27 +90,26 @@ export class ULWeb3Provider implements Provider {
...
-      this.uiController.showLocalStorageWarning();
-    } else {
-      this.walletService.loadFromStorage();
-    }
+  init() {
+    return this.initHandler.initialize();
   }
 
-  async init() {
+  private async _init() {
```
Marik-D: Do we use the pattern with underscores anywhere else in the project?

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -238,12 +234,15 @@ export class ULWeb3Provider implements Provider {
   }
 
   finalizeAndStop() {
+    return this.initHandler.finalize();
+  }
+
+  private _finalizeAndStop() {
```
Marik-D: Do we use the pattern with underscores anywhere else in the project?

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,39 @@
+type State<T, B> = {
+  kind: 'Initialized';
+  value: T;
+} | {
+  kind: 'Finalized';
```
Marik-D: Rename to stop

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,39 @@
+type State<T, B> = {
+  kind: 'Initialized';
```
Marik-D: I think names `InitializED` and `FinalizED` are a bit misleading because they imply that the process is already completed which might not be the case

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+  value: T;
+} | {
+  kind: 'Finalized';
+  value: B | undefined;
+};
+
+export class InitializationHandler<T, B> {
```
Marik-D: What happens if you call `initialize` and then immediately call `finalize` while the initialization hasn't completed yet?

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,68 @@
...
+        });
+      } else {
+        this.state = {kind: 'Finalized', value};
+      }
+    }
+    return this.state.value;
+  }
+}
+
+const isPromise = (value: unknown) => value && !!(value as any).then;
```
Marik-D:
```suggestion
const isPromise = (value: unknown): value is Promise<any> => value && !!(value as any).then;
```
This tells typescript that this function check for value to be a promise. It will also update the types if you use this in an if statement.

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,68 @@
+type State<T, B> = {
+  kind: 'None';
+} | {
+  kind: 'Initializing';
+  value: T | Promise<T>;
```
Marik-D: this should probably only be a promise

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,68 @@
...
+} | {
+  kind: 'Initialized';
+  value: T | Promise<T>;
+} | {
+  kind: 'Finalizing';
+  value: B | Promise<B>;
```
Marik-D: this should also only be a promise

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,56 @@
+type State<T, B> = {
+  kind: 'None';
+} | {
+  kind: 'Initializing';
+  value: Promise<T>;
+} | {
+  kind: 'Initialized';
+  value: Promise<T>;
```
Marik-D:
```suggestion
  value: T;
```
we have already awaited the value at this point

universal-login-commons/src/core/utils/InitializationHandler.ts
```diff
@@ -0,0 +1,56 @@
...
+} | {
+  kind: 'Finalizing';
+  value: Promise<B>;
+} | {
+  kind: 'Finalized';
+  value: Promise<B>;
```
Marik-D:
```suggestion
  value: B;
```

universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -238,12 +234,16 @@ export class ULWeb3Provider implements Provider {
...
+    this.walletService.stateProperty.set({kind: 'None'});
```
Marik-D: Is this needed?

Also setting the state property directly kinda breaks encapsulation. We should make the setter private so only `WalletService` can mutate it's state.

PR #2010, opened 2020-03-13
Add support for DAI

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -54,7 +56,7 @@ export const TopUpWithFiat = ({walletService, modalClassName, logoColor}: TopUpW
       return <Ramp
         address={contractAddress}
         amount={stringToEther(amount)}
```
xitronix: I suppose that If we support dai amount we can't just do stringToEther here

JustynaBroniszewska: Yes, the name can be confusing... I renamed it to `stringToWei`. Should be more descriptive now 😄

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -24,11 +24,13 @@ type TopUpWithFiatModal = 'none' | 'wait' | TopUpProvider;
...
   const [modal, setModal] = useState<TopUpWithFiatModal>('none');
   const [amount, setAmount] = useState('');
+  const [currency, setCurrency] = useState<string>('ETH');
   const contractAddress = walletService.getContractAddress();
   const relayerConfig = walletService.sdk.getRelayerConfig();
   const [paymentMethod, setPaymentMethod] = useState<TopUpProvider | undefined>(undefined);
 
-  const onPayClick = (provider: TopUpProvider) => {
+  const onPayClick = (provider: TopUpProvider, currency: string) => {
+    setCurrency(currency);
```
xitronix: I guess it should not be state. It can be just variable
I think this two line can run render 2 times
```
setCurrency(currency);
setModal(provider);
```

xitronix:
image

PR #2016, opened 2020-03-16
Enable changing header visibility in TopUpWithFiat component

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -51,6 +52,7 @@ export const TopUpWithFiat = ({walletService, modalClassName, logoColor}: TopUpW
         </>
       );
     case TopUpProvider.RAMP:
+      setHeaderVisible(false);
```
Marik-D: Side effects in render are bad practice

Marik-D: Instead do `useEffect` on `modal` and call it there

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -15,10 +15,11 @@ export interface TopUpWithFiatProps {
   walletService: WalletService;
   logoColor?: LogoColor;
   modalClassName?: string;
+  setHeaderVisible: (arg: boolean) => void;
```
Marik-D:
```suggestion
  setHeaderVisible: (isVisible: boolean) => void;
```

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
Marik-D: Also not sure if a callback to the parent component is the best solution here. Consider moving the state up and/or decomposing this component into two: `TopUpFiatDetails` and `TopUpWithOnRampProvider`

JustynaBroniszewska: Good point with `useEffect` - done

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
JustynaBroniszewska: done ✅

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -29,6 +30,14 @@ export const TopUpWithFiat = ({walletService, modalClassName, logoColor}: TopUpW
     setModal(provider);
   };
 
+  useEffect(() => {
+    if (modal === 'none') {
```
Marik-D: setHeaderVisble(modal === 'none')

Marik-D:
Or better, just move the `modal` state up, instead of `isHeaderVisible` it will be one less state that way

PR #2018, opened 2020-03-16
Introduce OnRampSuccessInfo

JustynaBroniszewska:


JustynaBroniszewska:
Screenshot 2020-03-16 at 13 43 00

universal-login-react/src/ui/TopUp/Fiat/TopUpWithFiat.tsx
```diff
@@ -78,11 +80,17 @@ export const TopUpWithFiat = ({setHeaderVisible, walletService, modalClassName,
         crypto="eth"
       />;
     case 'wait':
-      return <WaitingForOnRampProvider
-        onRampProviderName={TopUpProvider.RAMP}
-        className={modalClassName}
-        logoColor={logoColor}
-      />;
+      return walletService.isKind('Deployed')
```
xitronix: it should not be able to be here if wallet is Deployed.  It should redirect to /wallet ||  /creationSuccess (at least for wallet)

JustynaBroniszewska: It is possible because we have also regular TopUp in dashboard

PR #2026, opened 2020-03-17
Add missing dependencies and update

JustynaBroniszewska:


universal-login-web3/test/ULWeb3Provider.test.ts
```diff
@@ -76,7 +76,7 @@ describe('ULWeb3Provider', () => {
   describe('get accounts', () => {
     it('initialize onboarding when wallet does not exist', async () => {
       const initOnboardingSpy = sinon.spy(() => null);
-      sinon.replace(ulProvider, 'initOnboarding', initOnboardingSpy);
+      sinon.replace(ulProvider, 'initOnboarding', initOnboardingSpy as any);
```
Marik-D: Why is this needed?

xitronix: I resolved it. Initialization should return a Promise, not null

PR #2033, opened 2020-03-18
Add Jarvis Reward Token to wallet

JustynaBroniszewska:


universal-login-sdk/src/core/services/GasModeService.ts
```diff
@@ -5,6 +5,8 @@ import {GasPriceOracle} from '../../integration/ethereum/gasPriceOracle';
 import {PriceObserver} from '../observers/PriceObserver';
 import {GasPriceEstimation} from '../models/GasPriceSuggestion';
 
+const GAS_TOKENS = ['ETH', 'DAI'];
```
xitronix: I think we should split enves to GAS_TOKENS ANS TRANSER_TOKENS

JustynaBroniszewska: definitely

PR #2072, opened 2020-03-31
Distinguish wallet keys for different networks

xitronix:


universal-login-sdk/src/core/services/WalletStorageService.ts
```diff
@@ -1,36 +1,47 @@
...
-      const data = this.storageService.get(STORAGE_KEY);
-      if (data === null) {
-        return;
+      const data = this.storageService.get(DEPRECATED_STORAGE_KEY);
+      if (data === null) return;
+      const {wallet} = cast(JSON.parse(data), asObject({
+        kind: asExactly('Deployed'),
+        wallet: asApplicationWallet,
+      }));
+      for (const key of ['ropsten', 'rinkeby', 'kovan', 'mainnet']) {
```
JustynaBroniszewska, line 31: Extract `SUPPORTED_NETWORKS` or `SUPPORTED_CHAINS`

universal-login-sdk/src/core/services/WalletStorageService.ts
```diff
@@ -1,36 +1,47 @@
...
+        const network = key === 'mainnet' ? 'homestead' : key;
+        const provider = new providers.InfuraProvider(network);
+        const code = await provider.getCode(wallet.contractAddress);
```
JustynaBroniszewska, line 34: Maybe we should add new function to BlockchainService? 
```ts
isContract = async (address: string) =>
    '0x' !== await this.getCode(address)
```

then this code would look like: 
```ts
...
const blockchainService = new BlockchainService(new providers.InfuraProvider(network));
if (await blockchainService.isContract(wallet.contractAddress) {
...
```

universal-login-sdk/src/core/services/WalletStorageService.ts
```diff
@@ -1,36 +1,47 @@
...
-        return;
+      const data = this.storageService.get(DEPRECATED_STORAGE_KEY);
+      if (data === null) return;
+      const {wallet} = cast(JSON.parse(data), asObject({
+        kind: asExactly('Deployed'),
+        wallet: asApplicationWallet,
+      }));
+      for (const key of ['ropsten', 'rinkeby', 'kovan', 'mainnet']) {
+        const network = key === 'mainnet' ? 'homestead' : key;
+        const provider = new providers.InfuraProvider(network);
```
JustynaBroniszewska, line 33: And we can also extract function 
```ts
const getProviderForNetwork = (chainName: string) => {
    const network = key === 'mainnet' ? 'homestead' : key;
    return new providers.InfuraProvider(network);
}
```

JustynaBroniszewska: or maybee we can do something like this: 
```ts
const getNetworkWhereContractIsDeployed = async (address: string) => {
    for (const network of ['ropsten', 'rinkeby', 'kovan', 'mainnet']) {
       const provider = getProviderForNetwork(network);
       const blockchainService = new BlockchainService(provider);
           if (await blockchainService.isContract(address) {
               return network;
        }
    }
    return 'none'; // or null or whatever you want :) 
}
```

PR #2111, opened 2020-04-08
Reduce net_version calls to infura

xitronix:


universal-login-web3/src/ULWeb3Provider.ts
```diff
@@ -151,6 +153,9 @@ export class ULWeb3Provider implements Provider {
       case 'ul_disconnect':
         await this.finalizeAndStop();
         break;
+      case 'net_version':
+        const networkVersion = Network.toNumericId(this.network).toString();
+        return networkVersion;
```
Marik-D, line 158: Inline

PR #2114, opened 2020-04-09
Dashboard header redesign

alexcherman:
dashboard-mobile
dashboard-desktop


universal-login-react/src/ui/UFlow/DashboardModal.tsx
```diff
@@ -39,7 +39,7 @@ export const DashboardModal = ({walletService, onClose}: DashboardModalProps) =>
   };
 
   return (
-    <div className={`udashboard ${useThemeClassFor()}`}>
+    <div className={`udashboard ${useClassFor('udashboard')}`}>
```
JustynaBroniszewska, line 42: We should probably remove `udashobord` `unilogin-theme-.. unilogin-component-udashboard` is enough

alexcherman: Why should I remove ``unilogin-theme- ...``. We use this ``unilogin-theme -...+ unilogin-component -...`` combination in other components?

alexcherman: I'll remove ``dashboard`` class when redesign will be finished. Don't want to spend time renaming classes. I'll do it step by step and remove it when this class won't be needed.

JustynaBroniszewska: Ok 👍

PR #2138, opened 2020-04-20
Improve getFaviconUrl

xitronix:


universal-login-provider/src/applicationInfo.ts
```diff
@@ -6,17 +6,15 @@ export const getApplicationInfoFromDocument = () => {
...
-  let favicon;
   const nodeList = document.getElementsByTagName('link');
   for (let i = 0; i < nodeList.length; i++) {
     if ((nodeList[i].getAttribute('rel') === 'icon') || (nodeList[i].getAttribute('rel') === 'shortcut icon')) {
-      favicon = nodeList[i].getAttribute('href');
+      return nodeList[i].getAttribute('href');
     }
   }
-  return favicon;
 };
```
Sajma27, line 20: Shouldn't be returning null at the end, just in case?

universal-login-provider/src/applicationInfo.ts
```diff
@@ -6,17 +6,15 @@ export const getApplicationInfoFromDocument = () => {
 
 const getFaviconUrl = () => {
   const favicon = getFavicon();
-  const siteURL = document.URL;
+  const siteURL = document.URL.slice(0, -1);
```
JustynaBroniszewska, line 9: I would extract function `removeLastChar`

PR #2142, opened 2020-04-20
Transfer Jarvis's waiting-for styles to React

JustCaptcha:


universal-login-react/src/ui/styles/themes/Jarvis/waitingForThemeJarvis.sass
```diff
@@ -1,10 +1,16 @@
 .unilogin-theme-jarvis.unilogin-component-waitingfor
   .unilogin-component-waitingfor-modal-pending-section
+    margin: auto
     max-width: 570px
     padding: 10px 5px 30px
 
   .unilogin-component-waitingfor-action-description
     display: none
```
JustynaBroniszewska, line 8: Should it be here: `display:none` ?

xitronix:
image

xitronix:
On production: 
image

xitronix:
On production:

image

JustCaptcha:
image
image

JustynaBroniszewska:
I am not sure how it looks like now. If the last screenshots are correct, then it is wrong in WaitingForDeployment component. Because we have duplication of description. So we should show only one. I suggest the one on the bottom.

JustCaptcha:
image

PR #2145, opened 2020-04-20
Improve getting favicon's url

xitronix:


xitronix:
Tested on local kickback and gnosis-safe

universal-login-provider/src/applicationInfo.ts
```diff
@@ -6,10 +6,15 @@ export const getApplicationInfoFromDocument = () => {
...
 
+const getSiteUrl = (url: string) => url.split('#')[0];
+
 const getFaviconUrl = () => {
   const favicon = getFavicon();
-  const siteURL = removeLastChar(document.URL);
-  return favicon ? (siteURL + favicon) : null;
+  const siteURL = getSiteUrl(document.URL);
+  if (!favicon) return null;
+  return (siteURL[siteURL.length - 1] === '/' && favicon[0] === '/')
```
JustynaBroniszewska, line 15:
```suggestion
  if (!favicon) return null;
  const siteURL = getSiteUrl(document.URL);
  return (siteURL[siteURL.length - 1] === '/' && favicon[0] === '/')
```

PR #2162, opened 2020-04-23
Add type for DeviceType

JustynaBroniszewska:


universal-login-commons/src/core/constants/test.ts
```diff
@@ -40,7 +41,7 @@ export const TEST_EXECUTION_OPTIONS = {gasPrice: TEST_GAS_PRICE, gasLimit: TEST_
 export const TEST_APPLICATION_INFO = {
   applicationName: 'UniversalLogin',
   logo: 'logo',
-  type: 'laptop',
+  type: 'laptop' as DeviceType,
 };
```
xitronix, line 45: I think it might be better write something like this: 
```ts
export const TEST_APPLICATION_INFO: ApplicationInfo = {
  applicationName: 'UniversalLogin',
  logo: 'logo',
  type: 'laptop' as DeviceType,
}
```

universal-login-relayer/test/http/utils/getDeviceInfo.test.ts
```diff
@@ -15,7 +15,7 @@ describe('UNIT: getDeviceInfo', () => {
 
   it('iOS/Evidance', () => {
     req.useragent.source = 'deecert/4 CFNetwork/1098.7 Darwin/18.7.0';
-    const appInfo = {applicationName: 'Evidance', type: 'phone', logo: 'none'};
+    const appInfo = {applicationName: 'Evidance', type: 'phone' as DeviceType, logo: 'none'};
```
xitronix, line 18: It might be better to use type casting for appInfo layer?

PR #2175, opened 2020-04-27
Inject RefundPayerStore 

xitronix:


universal-login-relayer/src/core/services/execution/deployment/DeploymentHandler.ts
```diff
@@ -22,7 +24,7 @@ class DeploymentHandler {
       deviceInfo,
       state: 'Queued',
       contractAddress,
-      refundPayerId: apiKey && (await this.refundPayerValidator.store.get(apiKey))?.id,
+      refundPayerId: apiKey && (await this.refundPayerStore.get(apiKey))?.id,
```
JustynaBroniszewska, line 27: And now refundPayerStore can be private in RefundPayerValidator 😄

xitronix: Yeap, i forgot about it :D

PR #2210, opened 2020-04-30
Add GasUsed to messages and remove deployment sql repository

xitronix:


universal-login-relayer/src/integration/sql/services/SQLRepository.ts
```diff
@@ -60,9 +60,9 @@ export class SQLRepository<T extends Mineable> implements IRepository<T> {
   }
```
JustynaBroniszewska, line 60: We can  merge update functions in `markAsError` and `markAsPending` also

PR #2211, opened 2020-04-30
Setup test partner each time relayer starts

JustynaBroniszewska:


universal-login-relayer/src/http/relayers/RelayerUnderTest.ts
```diff
@@ -94,6 +94,11 @@ export class RelayerUnderTest extends Relayer {
     return this.config;
   }
 
+  async start() {
+    super.start();
```
JustynaBroniszewska, line 98:
```suggestion
    await super.start();
```

PR #2213, opened 2020-04-30
Throw if address is not a contract

JustynaBroniszewska:


universal-login-contracts/src/integration/BlockchainService.ts
```diff
@@ -68,6 +68,7 @@ export class BlockchainService {
 
   async fetchProxyVersion(contractAddress: string) {
     const proxyBytecode = await this.getCode(contractAddress);
+    ensure(proxyBytecode !== '0x', Error, 'Address is not a contract');
```
xitronix, line 71: Use InvalidContract Error

PR #2254, opened 2020-05-07
Introduce ChooseTopUpCryptoCoin modal

JustCaptcha:


JustCaptcha:
image

image

universal-login-react/src/ui/TopUp/ChooseTopUpToken.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <div className={classForComponent('top-up-header')}>
+        <div className={classForComponent('top-up-crypto-coin')}>
+          <ModalTitle>How do you wanna pay for account creation?</ModalTitle>
+          <div className={classForComponent('top-up-methods')}>
+            {supportedTokens && supportedTokens.map((token) => {
+              return (
+                <div key={token} className={`${classForComponent('top-up-method')}`}>
+                  <div className={classForComponent('top-up-radio-inner')}>
+                    <div className={classForComponent('top-up-method-icons')}>
+                      <img className={classForComponent('top-up-method-icon')} src={token === 'ETH' ? EthereumIcon : DaiIcon} alt={token} />
```
JustynaBroniszewska, line 29: Let's add some button here and callback also. This callback could be passed in the constructor and for now, it can be just simple `console.log`
```ts
(token) => console.log(token);
```

JustynaBroniszewska: Btw we can use `getIconForToken` function. It returns a proper image based on the token.

universal-login-react/src/ui/TopUp/TopUpRadio.tsx
```diff
@@ -36,7 +37,7 @@ export const TopUpRadio = ({id, onClick, checked, children, className, name}: To
 export const TopUpRadioCrypto = (props: Omit<TopUpRadioProps, 'children'>) => (
   <TopUpRadio {...props}>
     <div className={classForComponent('top-up-method-icons')}>
-      <img className={classForComponent('top-up-method-icon')} src={EthereumIcon} alt="Ethereum" />
+      <img className={classForComponent('top-up-method-icon')} src={props.iconCrypto || EthereumIcon} alt="Ethereum" />
```
JustynaBroniszewska, line 40: Why aren't we using `getIconForToken` here also?

PR #2276, opened 2020-05-12
Remove unnecessary wrapper from Transfer

JustCaptcha:
image


JustynaBroniszewska:
If we don't use universal-login-default className anywhere, we should probably remove also this:

```css

  .universal-login-default
    flex: 1
```

PR #2285, opened 2020-05-13
Adjust TopUp component to refund with token

JustynaBroniszewska:


universal-login-react/src/ui/TopUp/ChooseTopUpMethodHeader.tsx
```diff
@@ -8,9 +8,10 @@ import {classForComponent} from '../utils/classFor';
 interface ChooseTopUpMethodHeaderProps {
   topUpMethod?: TopUpMethod;
   setTopUpMethod: (arg: TopUpMethod) => void;
+  topUpCurrency: string;
 };
```
xitronix, line 12: Can we use ChooseTopUpMethodProps here?

universal-login-react/src/ui/TopUp/TopUp.tsx
```diff
@@ -23,11 +23,13 @@ export interface TopUpProps {
 export const TopUp = ({walletService, modalClassName, hideModal, isModal, logoColor}: TopUpProps) => {
   const [topUpMethod, setTopUpMethod] = useState<TopUpMethod>(undefined);
   const [headerVisible, setHeaderVisible] = useState<boolean>(true);
+  const topUpCurrency = walletService.isKind('Future') ? walletService.getFutureWallet().getTopUpCurrencySymbol() : 'ETH or DAI';
```
xitronix, line 26: What if we want use not only ETH or DAI in regular TopUp?

xitronix:
image
image
minimal amount for DAI and ETH is same

PR #2298, opened 2020-05-15
Remove .universal-login-default selectors

JustCaptcha:


universal-login-react/src/ui/Modals/ModalWrapper.tsx
```diff
@@ -29,7 +29,7 @@ export const ModalWrapper = ({children, modalClassName, hideModal, message}: Mod
 
   return (
     <div className={useClassFor('modal-backdrop')}>
-      <div className={modalClassName ? `universal-login ${modalClassName}` : 'universal-login universal-login-default'}>
+      <div className={modalClassName ? `universal-login ${modalClassName}` : 'universal-login'}>
```
JustynaBroniszewska, line 32: Do we use modalClassName anywhere?

universal-login-react/src/ui/styles/themes/Jarvis/modalThemeJarvis.sass
```diff
@@ -68,9 +68,6 @@
     color: #FFFFFF
     line-height: 19px
 
-  .universal-login-default
-    height: 100%
-
     @media only screen and (max-width: 770px)
```
JustynaBroniszewska, line 71: Shouldn't we remove it also?

universal-login-react/src/ui/Modals/ModalWrapper.tsx
JustCaptcha: Yes, we are using it in LocalStorageBlockedWarningScreen, DashboardModal.

universal-login-react/src/ui/styles/themes/Jarvis/modalThemeJarvis.sass
JustCaptcha: I actually removed it.

JustynaBroniszewska: No, `@media only ...` is not removed 😉

PR #2312, opened 2020-05-19
Validate message

xitronix:


universal-login-sdk/test/api/wallet/DeployedWallet.int.test.ts
```diff
@@ -56,13 +55,13 @@ describe('INT: DeployedWallet', () => {
   });
 
   it('setRequiredSignatures', async function () {
-    await waitForSuccess(deployedWallet.addKey(publicKey, {gasPrice, gasToken: mockToken.address}));
```
JustynaBroniszewska, line 59: Why do we want to remove gasToken? Maybe we can change gasPrice to `TEST_GAS_PRICE_IN_TOKEN`?

PR #2331, opened 2020-05-22
Add UniLogin theme styles for Udashboard's navigation on mobile

JustCaptcha:
image

### DIFF
--- a/universal-login-react/src/ui/styles/themes/UniLogin/noticeThemeUniLogin.sass
+++ b/universal-login-react/src/ui/styles/themes/UniLogin/noticeThemeUniLogin.sass
@@ -1,0 +1,8 @@
+.unilogin-theme-unilogin.unilogin-component-notice-wrapper
+  background-color: #fff
+
+  &:before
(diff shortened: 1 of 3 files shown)


universal-login-react/src/ui/styles/themes/UniLogin/noticeThemeUniLogin.sass
```diff
@@ -0,0 +1,11 @@
...
+  @media only screen and (max-width: 600px)
+    background-color: #fff
+    position: absolute
+    bottom: 0
+    background: #8884A7
+    opacity: 1
+
+  &:before
+    @media only screen and (max-width: 600px)
+      background-color: #fff
```
JustynaBroniszewska, line 11: Do we need these few lines? Background-color is already set to #fff

JustCaptcha: Yes, because we should change :before's background-color to white as well.

PR #2344, opened 2020-05-29
Correctly show usd value

xitronix:


universal-login-commons/src/core/utils/safeMultiply.ts
```diff
@@ -8,14 +8,21 @@ export const safeMultiplyAndFormatEther = (valueInWei: utils.BigNumber, multipli
...
   const normalizedMultiplier = utils.parseEther(multiplier.toString());
   const resultE18 = normalizedMultiplier.mul(value);
   const result = utils.formatEther(resultE18);
   return cutPrecisionForInteger(result);
 };
 
-const cutPrecisionForInteger = (value: string) => value.split('.')[1] === '0' ? value.split('.')[0] : value;
+const cutPrecisionForInteger = (value: string) => value.split('.')[1] === '0' ? cutPrecision(value) : value;
+
+const cutPrecision = (value: string) => value.split('.')[0];
```
JustynaBroniszewska, line 25: What if the first number after `.` will be 9?

PR #2353, opened 2020-06-17
Reduce requests to infura by check only logs for types with subscriptions

xitronix:


universal-login-sdk/src/core/observers/WalletEventsObserver.ts
```diff
@@ -23,7 +23,8 @@ export class WalletEventsObserver {
   }
 
   async fetchEvents(lastBlock: number, types: WalletEventType[]) {
-    for (const type of types) {
+    const typesWithSubscriptions = types.filter(type => this.observableRecords[type].length > 0);
```
JustynaBroniszewska, line 26: Can we extract it somehow? `getSubscribedEvents` or something that will show what happens ehre

xitronix: I suppose the name of the variable shows what happens here?

PR #2360, opened 2020-06-23
Extract CoingeckoApi

xitronix:


universal-login-commons/src/core/services/TokenPricesService.ts
```diff
@@ -1,15 +1,13 @@
...
+import {CoingeckoApi, TokenDetailsWithCoingeckoId} from '../../integration/http/CoingeckoApi';
 const cryptocompare = require('cryptocompare');
-import {Sanitizer, asObject, asNumber, cast} from '@restless/sanitizers';
-
-interface TokenDetailsWithCoingeckoId extends TokenDetails {
-  coingeckoId: string;
-};
 
 export class TokenPricesService {
+  private coingeckoApi = new CoingeckoApi();
```
JustynaBroniszewska, line 6: We should inject it in the constructor

PR #2373, opened 2020-06-26
Make gasLimit optional in GasPrice component

JustynaBroniszewska:


universal-login-react/src/ui/commons/GasPrice/GasPrice.tsx
```diff
@@ -101,15 +101,15 @@ export const GasPriceWithOptions = ({isDeployed = true, deployedWallet, sdk, gas
...
+            {gasLimit ? <TransactionFeeChoose
               gasModes={gasModes}
               modeName={modeName}
               tokenAddress={gasOption.token.address}
               gasLimit={gasLimit}
               usdAmount={usdAmount}
               tokensDetailsWithBalance={tokenDetailsWithBalance}
               onGasOptionChanged={onGasOptionSelected}
-            />
+            /> : <Spinner/>}
```
xitronix, line 112: Why do we show spinner here instead of the 121st line

PR #2397, opened 2020-07-08
Introduce EmailConfirmationHandler

xitronix:


universal-login-relayer/src/http/relayers/Relayer.ts
```diff
@@ -120,6 +125,9 @@ class Relayer {
     const devicesStore = new DevicesStore(this.database);
     const devicesService = new DevicesService(devicesStore, relayerRequestSignatureValidator);
     const futureWalletStore = new FutureWalletStore(this.database);
+    const emailConfirmationStore = new EmailConfirmationsStore(this.database);
+    this.emailService = new EmailService(this.config.emailAddress, this.config.emailPassword);
+    const emailConfirmationHandler = new EmailConfirmationHandler(emailConfirmationStore, this.emailService);
```
JustynaBroniszewska, line 130: Probably emailConfirmationHandler should be also a property of the relayer, ether way assigning empty function to `sendConfirmationMail` function won't work for it.

xitronix: For this purpose, I have added emailService as a property

PR #2415, opened 2020-07-14
Improve EstimateGasValidator

xitronix:


universal-login-relayer/test/integration/ethereum/validators/MessageExecutionValidator.test.ts
```diff
@@ -31,11 +30,6 @@ describe('INT: MessageExecutionValidator', () => {
     await expect(messageExecutionValidator.validate(signedMessage)).to.not.be.rejected;
   });
```
JustynaBroniszewska, line 32: why did we remove this test?

xitronix: Because we removed EstimateGasValidator from MessageExecutionValidator

PR #2422, opened 2020-07-15
Add validatePassword

Sajma27:
# Summary

Fixes #number

## Checklist

### DIFF
--- a/universal-login-commons/src/core/utils/isProperPassword.ts
+++ b/universal-login-commons/src/core/utils/isProperPassword.ts
@@ -1,11 +1,9 @@
 const minLength = 10;
 const minNumberOfCapitals = 1;
-const minNumberOfSpecials = 1;
-const minNumberOfDigits = 1;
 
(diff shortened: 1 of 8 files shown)


universal-login-commons/src/core/utils/isProperPassword.ts
```diff
@@ -0,0 +1,11 @@
+const minLength = 10;
+const minNumberOfCapitals = 1;
+const minNumberOfSpecials = 1;
+const minNumberOfDigits = 1;
+
+export const isProperPassword = (password: string) => {
+  const regExp = new RegExp(
```
JustynaBroniszewska, line 7: I think that we should check only for min length and capital letters usage. Special symbols and digits should be optional. To be honest, I find it very frustrating when someone requires a special character in a password. So, a password, in my opinion, should be min 10-symbols long and has min 1 capital letter.

Sajma27: Done

PR #2424, opened 2020-07-16
Improve TokenDetailsStore.getTokenBy

xitronix:


universal-login-sdk/test/core/services/TokensDetailsStore.test.ts
```diff
@@ -48,6 +48,7 @@ describe('INT: TokensDetailsStore', () => {
     const expectedToken = tokensDetailsStore.tokensDetails[0];
 
     expect(tokensDetailsStore.getTokenBy('symbol', expectedToken.symbol)).be.deep.eq(expectedToken);
+    expect(tokensDetailsStore.getTokenBy('symbol', expectedToken.symbol.toLowerCase())).be.deep.eq(expectedToken);
```
sz-piotr, line 51: lowercasing here defeats the point of the test. You should pass the value as mixed case and expect it to still come out as stored

universal-login-sdk/test/core/services/TokensDetailsStore.test.ts
```diff
@@ -60,6 +61,7 @@ describe('INT: TokensDetailsStore', () => {
     await tokensDetailsStore.fetchTokensDetails();
     const expectedToken = tokensDetailsStore.tokensDetails[0];
     expect(tokensDetailsStore.getTokenBy('address', expectedToken.address)).to.deep.eq(expectedToken);
+    expect(tokensDetailsStore.getTokenBy('address', expectedToken.address.toLowerCase())).to.deep.eq(expectedToken);
```
sz-piotr, line 64: same here

PR #2425, opened 2020-07-16
Introduce top-up with usdc

xitronix:


universal-login-react/src/core/utils/formatValue.ts
```diff
@@ -0,0 +1,5 @@
+import {utils} from 'ethers';
+import UniLoginSdk from '@unilogin/sdk';
+
+export const formatValue = (sdk: UniLoginSdk, amount: string, currency: string) =>
```
JustynaBroniszewska, line 4: it should get `(amount, currency, decimals)`

JustynaBroniszewska: then testing will be much more easier

xitronix: The intension was easy to use. Testing wasn't difficult

PR #2449, opened 2020-07-24
Introduce isButtonDisabled function

JustynaBroniszewska:


universal-login-commons/src/core/utils/isProperPassword.ts
```diff
@@ -1,9 +1,9 @@
 const minLength = 10;
 const minNumberOfCapitals = 1;
 
-export const isProperPassword = (password: string) => {
+export const isProperPassword = (password?: string) => {
```
xitronix, line 4: should be optional?

PR #2454, opened 2020-07-24
Remove username suggestions field from EmailFlowChooser

JustCaptcha:
image


JustynaBroniszewska:
`YOUR EMAIL` label is stuck to the first input. I don't think this was the intention

PR #2455, opened 2020-07-24
Decrease error-hint's padding-left value

JustCaptcha:
image
image


JustynaBroniszewska:
Could you also add styles for hints in EnterPassword component?

PR #2468, opened 2020-07-29
Add styles for Confirm code modal

JustCaptcha:
image
image
image
image


universal-login-react/src/ui/Onboarding/ReactCodeInput.tsx
```diff
@@ -345,30 +351,32 @@ export class ReactCodeInput extends Component<ReactCodeInputProps, any> {
...
-              onFocus={(e) => e.target.select()}
-              onBlur={(e) => this.handleBlur(e)}
-              onChange={(e) => this.handleChange(e)}
-              onKeyDown={(e) => this.handleKeyDown(e)}
-              disabled={disabled}
-              data-valid={isValid}
-              pattern={pattern}
-              inputMode={inputMode}
-            />
+            <div className={`${wrapperClassName || 'react-code-wrapper'} ${value && 'it-has-value'}`} style={{display: 'inline-block'}} key={`input_${i}`}>
```
JustynaBroniszewska, line 354: Key should be named: `input_wrapper_${i}`
We should also move style to sass file (IMO it shouldn't be here)
Why do we want to use a different name? I would just add className `react-code-input-wrapper` and add style for that.

universal-login-react/src/ui/Onboarding/ReactCodeInput.tsx
```diff
@@ -345,30 +351,32 @@ export class ReactCodeInput extends Component<ReactCodeInputProps, any> {
...
+                id={`${this.uuid}-${i}`}
+                data-id={i}
+                autoFocus={autoFocus && i === 0}
+                value={value}
+                type={type}
+                min={0}
+                max={9}
+                maxLength={input.length === i + 1 ? 1 : input.length}
+                className={inputClassName}
+                style={inputClassName !== '' ? {} : styles.input}
```
JustynaBroniszewska, line 368: We can simplify it: `style={inputClassName || styles.input}`

JustCaptcha: This case don't work, because style can't be undefined. I improved this line in this way:
```style={inputClassName ? styles.input : {}}```

PR #2508, opened 2020-08-12
Split EncryptedWallet and StoredEncryptedWallet models

JustynaBroniszewska:


universal-login-commons/src/core/models/wallets/EncryptedWallet.ts
```diff
@@ -1,5 +1,32 @@
 import {asObject, asString, asNumber} from '@restless/sanitizers';
-import {StoredEncryptedWallet, EncryptedWallet, Crypto, CipherParams, KdfParams} from '../../models/wallets/StoredEncryptedWallet';
+
+export interface KdfParams {
```
xitronix, line 3: remove export for local interfaces and sanitizers

JustynaBroniszewska: ok

PR #2512, opened 2020-08-12
Introduce RequestedRestoring wallet

xitronix:


universal-login-sdk/src/core/services/WalletSerializer.ts
```diff
@@ -22,6 +22,13 @@ export class WalletSerializer {
             ensName: state.wallet.ensName,
           },
         };
+      case 'RequestedRestoring':
```
JustynaBroniszewska, line 25: Let's add deserialize also.

PR #2523, opened 2020-08-17
Improve AggregateBalanceObserver

xitronix:


universal-login-commons/test/core/services/TokensValueConverter.test.ts
```diff
@@ -13,13 +13,24 @@ describe('UNIT: TokensValueConverter', () => {
...
+
+      expect(actualEthTotalWorth).to.deep.eq({
+        USD: 2 * currencyToValue.USD,
+        DAI: 2 * currencyToValue.DAI,
+        SAI: 2 * currencyToValue.SAI,
+        ETH: 2 * currencyToValue.ETH,
+      });
+    });
+
+    it('2 ETH with 6 decimals', () => {
```
JustynaBroniszewska, line 32: It is very confusing because we all know that ether has 18 decimals. We should change the whole test to 'mock token'. So it would be something like: 
```ts
it('2 MT (mock token) with 6 decimals', () => {
   const actualEthTotalWorth = tokensValueConverter.getTokenTotalWorth({balance: utils.parseUnits('2', 6), symbol: 'MT', address: TEST_MOCK_TOKEN, decimals: 6}, currencyToValue);
}
```

xitronix: Does it important? Decimal in the token matter, not name or symbol

PR #2527, opened 2020-08-17
Make encoding transfer amount in token depended on decimals

xitronix:


universal-login-commons/src/core/utils/sanitizers/asTransferDetails.ts
```diff
@@ -1,10 +1,11 @@
-import {asObject, asString, Sanitizer} from '@restless/sanitizers';
+import {asObject, asString, Sanitizer, asNumber} from '@restless/sanitizers';
 import {TransferDetails} from '../../..';
 import {asGasParameters} from './asGasParameters';
 
 export const asTransferDetails: Sanitizer<TransferDetails> = asObject<TransferDetails>({
   to: asString,
   amount: asString,
   transferToken: asString,
+  decimals: asNumber,
```
JustynaBroniszewska, line 9: It looks pretty bad... I don't like the idea that `decimals` is one of the `TransferDetails` property. It should be a token property, so we would have something like this:
```ts 
    interface TransferDetails {
        ...
        token: Token
        ...
    }
    interface Token {
        address: string;
        decimals: number;
    }
```

I think that the perfect idea would be to use CurrencyValue (but first, we would need to add property `decimals` to it). I know that this a big refactoring, but I think it is worth trying.

universal-login-commons/src/core/models/transactions.ts
```diff
@@ -1,10 +1,13 @@
 import {utils} from 'ethers';
 import {GasParameters} from './gas';
+import {TokenDetails} from './TokenData';
+
+export type TransferToken = Pick<TokenDetails, 'address' | 'decimals'>;
```
sz-piotr, line 5: why not declare the interface explicitly?

universal-login-react/src/ui/Transfer/Transfer.tsx
```diff
@@ -22,20 +22,19 @@ export interface TransferProps {
 }
 
 export const Transfer = ({transferService, onTransferTriggered, sdk}: TransferProps) => {
-  const [transferDetails, setTransferDetails] = useState<TransferDetails>({transferToken: ETHER_NATIVE_TOKEN.address, amount: '', to: '', gasParameters: INITIAL_GAS_PARAMETERS});
+  const [transferDetails, setTransferDetails] = useState<TransferDetails>({token: {address: ETHER_NATIVE_TOKEN.address, decimals: 0} , amount: '', to: '', gasParameters: INITIAL_GAS_PARAMETERS});
```
sz-piotr, line 25: shouldn't the default be 18 not 0?

PR #2549, opened 2020-08-25
Introduce DeployedWithoutEmail state

xitronix:


universal-login-sdk/src/core/services/WalletStorageService.ts
```diff
@@ -22,22 +21,20 @@ export class WalletStorageService implements WalletStorage {
...
-      for (const network of ['mainnet', 'kovan', 'rinkeby', 'ropsten']) {
-        const providerService = new ProviderService(getProviderForNetwork(network));
-        if (await providerService.isContract(wallet.contractAddress)) {
-          this.storageService.remove(DEPRECATED_STORAGE_KEY);
-          this.storageService.set(`${DEPRECATED_STORAGE_KEY}-${network}`, data);
-          return;
+    for (const network of ['mainnet', 'kovan', 'rinkeby', 'ropsten']) {
+      try {
+        const data = this.storageService.get(`${DEPRECATED_STORAGE_KEY}-${network}`);
+        if (data === null) continue;
```
JustynaBroniszewska, line 27: Why we should continue? I believe we should return here

universal-login-sdk/src/core/services/WalletStorageService.ts
```diff
@@ -22,22 +21,20 @@ export class WalletStorageService implements WalletStorage {
...
+        const state = cast(JSON.parse(data), asObject({
+          kind: asExactly('Deployed'),
+          wallet: asSerializedDeployedWallet,
+        }));
+        if (state?.kind !== 'Deployed') continue;
```
JustynaBroniszewska, line 32: We should also return here

JustynaBroniszewska: But I also think that this situation is not possible, because we are using cast a few lines higher. If cast won't get `Deployed` wallet it will throw an error and we catch it.

universal-login-sdk/src/core/services/WalletStorageService.ts
xitronix: Because we should migrate wallet for every network if we have such possibility. So if data is null we continue migrate other networks

universal-login-sdk/src/core/services/WalletStorageService.ts
xitronix: 👍 I forgot why I added it, but I think is possible because of typescript. I will check it

universal-login-sdk/src/core/services/WalletStorageService.ts
JustynaBroniszewska: ok

PR #2557, opened 2020-08-26
Introduce RequestedMigratingWallet

JustynaBroniszewska:


universal-login-sdk/src/api/wallet/RequestedMigrating.ts
```diff
@@ -0,0 +1,30 @@
+import {SerializableRequestedMigratingWallet} from '@unilogin/commons';
+import UniLoginSdk from '../sdk';
+
+export class RequestedMigratingWallet implements SerializableRequestedMigratingWallet {
```
xitronix, line 4: Shouldn't also extends deployedWallet?

universal-login-sdk/src/api/wallet/RequestedMigrating.ts
```diff
@@ -0,0 +1,33 @@
...
+export class RequestedMigratingWallet extends DeployedWallet implements SerializableRequestedMigratingWallet {
+  constructor(
+    readonly contractAddress: string,
+    readonly ensName: string,
+    readonly privateKey: string,
+    readonly email: string,
+    public readonly sdk: UniLoginSdk,
+  ) {
+    super(contractAddress, ensName, privateKey, sdk, email);
+  };
```
xitronix, line 14: Is it necessary? I think if we don't declare the constructor, we will be able to use DeployedWallet constructor from RequestedMigratingWallet level?

xitronix (approved):

PR #2568, opened 2020-09-02
Introduce EnterEmail component

xitronix:
image

### DIFF
--- a/universal-login-react/src/ui/Migrating/EnterEmail.tsx
+++ b/universal-login-react/src/ui/Migrating/EnterEmail.tsx
@@ -1,5 +1,6 @@
 import React, {useState} from 'react';
-import {ensureNotFalsy, isProperEmail} from '@unilogin/commons';
+import {useHistory} from 'react-router';
+import {ensureNotFalsy} from '@unilogin/commons';
 import {WalletService} from '@unilogin/sdk';
 import {useClassFor, classForComponent} from '../utils/classFor';
 import {InputField} from '../commons/InputField';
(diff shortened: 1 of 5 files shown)


universal-login-react/src/ui/Migrating/EnterEmail.tsx
```diff
@@ -0,0 +1,77 @@
...
+import {PrimaryButton} from '../commons/Buttons/PrimaryButton';
+import '../styles/base/enterEmail.sass';
+import '../styles/themes/UniLogin/enterEmailThemeUniLogin.sass';
+import emailIcon from '../assets/icons/e-mail.svg';
+import {SecondaryButton} from '../commons/Buttons/SecondaryButton';
+import Spinner from '../commons/Spinner';
+import {OnboardingStepsWrapper} from '../Onboarding/OnboardingStepsWrapper';
+import {useHistory} from 'react-router';
+
+const emailValidator = {
```
JustynaBroniszewska, line 16: We can extract it to separate file e.g. `inputValidators` and get rid of duplications

universal-login-react/src/ui/Migrating/EnterEmail.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export const EnterEmail = ({hideModal, walletService, onConfirm}: EnterEmailProps) => {
+  const history = useHistory();
+  const [email, setEmail, emailError] = useInputField([emailValidator]);
+  const [loading, setLoading] = useState(false);
+  const primaryButtonClassName = useClassFor('proceed-btn');
+
+  const onConfirmClick = async () => {
+    ensureNotFalsy(email, Error, 'Email missing');
```
JustynaBroniszewska, line 34: `Email is missing`

universal-login-react/src/ui/Migrating/EnterEmail.tsx
```diff
@@ -0,0 +1,77 @@
...
+    setLoading(true);
+    await onConfirm(email);
```
JustynaBroniszewska, line 36: Is it possible that onConfirm callback will be not provided or won't change the modal state? If so, we could setLoading to false after calling onConfirm

xitronix: No, the component doesn't have sens if we are not changing screen

JustynaBroniszewska (approved):

PR #2571, opened 2020-09-02
Extract ensName and email in WalletSerializer test

JustynaBroniszewska:


universal-login-sdk/test/core/services/WalletSerializer.test.ts
```diff
@@ -9,6 +9,9 @@ import {RestoringWallet} from '../../../src/api/wallet/RestoringWallet';
 import {RequestedRestoringWallet} from '../../../src/api/wallet/RequestedRestoringWallet';
 
 describe('UNIT: WalletSerializer', () => {
+  const ensName = 'name.mylogin.eth';
```
xitronix, line 12: So we can extract it to TEST_ENS_NAME and use it everywhere. Same with email

xitronix (approved):

PR #2598, opened 2020-09-14
Change coingeckoApi for relayer api

xitronix:


universal-login-commons/src/integration/http/Coingecko.ts
```diff
@@ -0,0 +1,49 @@
+import {Sanitizer, asObject} from '@restless/sanitizers';
+import {} from './http';
```
JustynaBroniszewska, line 2: Is it important?

JustynaBroniszewska (approved):

